PR #1, opened 2017-04-22
Fixes/line height consistency

danwalmsley:
This fixes the inconsistencies between empty lines and lines with at least 1 character in.


aelij:
Nice find :)

PR #4, opened 2017-04-22
Features/build scripts

danwalmsley:
Work in progress, don't merge just yet!

This adds the following:

- .net 4.6 specific csproj.
- .net core 1.1 project
- fixes release builds.
- adds cake build script to build, test and package nugets (when unit tests and nuget packages exist.)
- adds appveyor and travis scripts, so we will be able to test in cloud on windows, Linux and mac.
- adds appveyor and travis status badges to the readme.
- adds codefactor badge to the readme.

(see preview of readme here) https://github.com/danwalmsley/AvaloniaEdit/tree/features/build-scripts
We will need to set appveyor and travis and code factor up on your main repo, and I need to edit build.cake to point to that.

However we may want to have the main repo part of Avalonia?

Let me know!


aelij:
* I'm almost sure there's no need to separate projects for 4.6/Core these days. But if they are needed, use `<PackageReference>`s instead of packages.config - they also work in the "old" csproj
* Why is cake needed? Doesn't MSBuild/dotnet CLI already do all that for netstandard libraries?
* I've asked grokys about hosting this under the Avalonia org. He said to publish it first under my account, then we'll transfer it. Perhaps we should hold off with CI until this is done?

danwalmsley:
Cake gives us a cross-platform way of publishing nugets, enumerating projects to build, zipping the artifacts, if you don't use cake we need separate build scripts for Linux and windows and mac, avalonia, itself is using it, I think we should get it set up asap so we can start getting tests running.

danwalmsley:
@aelij I think this is ready to merge now once the ci completes, if your happy,

It now uses vs2017 csproj format and builds both netcore and net46, from single csproj.

src/AvaloniaEdit/Rendering/VisualLine.cs
```diff
@@ -375,11 +375,11 @@ public double GetTextLineVisualYPosition(TextLine textLine, VisualYPosition yPos
                         case VisualYPosition.LineBottom:
                             return pos + tl.Height;
```
aelij: Could you undo this change? I'm working on a different fix.

PR #6, opened 2017-04-23
Feature/ilogical scrolling (fixes #2)

### LINKED ISSUE #2: Handle scrolling
Most of the scrolling code is currently commented out since Avalonia doesn't have `IScrollInfo`. Possibly `ILogicalScrollable` could be made to work here.

Most of the code is in `TextView` (I deleted much of the scroll code, may need to copy again from AvalonEdit).

danwalmsley:
Saved you a massive headache there!

Working scrolling!

### DIFF
--- a/src/AvaloniaEdit/Editing/TextArea.cs
+++ b/src/AvaloniaEdit/Editing/TextArea.cs
@@ -1049,13 +1049,25 @@ namespace AvaloniaEdit.Editing
 
         public IList<RoutedCommandBinding> CommandBindings { get; } = new List<RoutedCommandBinding>();
 
-        public bool IsLogicalScrollEnabled => true;
+        public bool IsLogicalScrollEnabled { get; } = true;
 
         public Action InvalidateScroll { get; set; }
 
-        public Size ScrollSize => new Size(Bounds.Width, 2);
+        public Size ScrollSize
+        {
+            get
+            {
+                return new Size(Bounds.Width, 2);
+            }
+        }
 
-        public Size PageScrollSize => throw new NotImplementedException();
+        public Size PageScrollSize
+        {
+            get
+            {
+                throw new NotImplementedException();
+            }
+        }
 
         public Size Extent { get; private set; }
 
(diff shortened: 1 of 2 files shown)


src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -936,7 +972,104 @@ internal void OnTextCopied(TextEventArgs e)
             TextCopied?.Invoke(this, e);
         }
 
+        private int LogicalScrollSize
+        {
+            get
+            {
+                if (TextView?.Document != null)
+                {
+                    return TextView.Document.LineCount + 2;
```
aelij: `2` should probably be a constant (e.g. `MinimumLineCount`)

src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -936,7 +972,104 @@ internal void OnTextCopied(TextEventArgs e)
...
+        }
+
+        public IControl GetControlInDirection(NavigationDirection direction, IControl from)
+        {
+            return null;
+        }
+
         public IList<RoutedCommandBinding> CommandBindings { get; } = new List<RoutedCommandBinding>();
+
+        public bool IsLogicalScrollEnabled => true;
```
aelij: I think `ILogicalScrollable` members should be implemented explicitly

aelij:
Great! Thanks :)

src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -1049,13 +1049,25 @@ public IControl GetControlInDirection(NavigationDirection direction, IControl fr
 
         public IList<RoutedCommandBinding> CommandBindings { get; } = new List<RoutedCommandBinding>();
 
-        public bool IsLogicalScrollEnabled => true;
+        public bool IsLogicalScrollEnabled { get; } = true;
```
aelij: Oh, not what I meant. The expression syntax is preferred, so undo these changes.

I meant that all of the `ILogicalScrollable` members should not be public. So instead of `public bool IsLogicalScrollEnabled => true`, use `bool ILogicalScrollable.IsLogicalScrollEnabled => true`

danwalmsley: Ah I see what you mean, yeah I agree that is better! :)

aelij:
Great!

PR #7, opened 2017-04-23
render caret the correct size and at correct location.

danwalmsley:
This raises the question:

Do you know what HeightFactor and BaselineFactor are?

### DIFF
--- a/src/AvaloniaEdit/Rendering/VisualLine.cs
+++ b/src/AvaloniaEdit/Rendering/VisualLine.cs
@@ -375,11 +375,11 @@ namespace AvaloniaEdit.Rendering
                         case VisualYPosition.LineBottom:
                             return pos + tl.Height;
                         case VisualYPosition.TextTop:
-                            return pos + tl.Baseline - _textView.DefaultBaseline;
+                            return pos + tl.Baseline - (_textView.DefaultBaseline * TextLineRun.BaselineFactor);
                         case VisualYPosition.TextBottom:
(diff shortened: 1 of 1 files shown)


aelij:
> Do you know what HeightFactor and BaselineFactor are?
These are approximations I made since there's no text metrics API in Avalonia. They are actually derived from each typeface (font family).

I'm not sure multiplying by these factors is the right fix. `CalculateDefaultTextMetrics` should have produced the correct results. These constants should not be used outside of the `TextLine` implementation.

aelij:
Baseline

aelij:
Hmm, seems like there's a major difference between WPF's text rendering and Avalonia's. WPF's `DrawingContext` draws glyph runs using the Baseline as the Y coordinate, while Avalonia uses the top-left coordinate. Perhaps all of these baseline calculations aren't needed at all.

danwalmsley:
@aelij I'm pretty sure formatted text class takes care of this for us but might be wrong.

danwalmsley:
Lets resolve if we need the factors in another pr.

aelij:
It doesn't. Imagine you have a text line with all sorts of fonts. `FormattedText` could represent a single text run within that line. Different fonts have different baselines (and line heights) so they need to be aligned. And that's what `TextLine` essentially does - it creates runs and aligns them in a line.

While this is not a concern for AvaloniaEdit (since it only allows a single font), if in the future we'd like to have a rich text box in Avalonia the `TextLine` implementation here could potentially be used.

For the time being, I'm going to close this PR and resolve this in another way (already have an idea how).

PR #12, opened 2017-04-30
Completion window

aelij:
Fixes #3

PR #14, opened 2017-05-01
Insight window

aelij:
Fixes #12


danwalmsley:
@aelij are you still getting black boxes I still get black boxes on this branch when pressing '.'

Also how can I activate the insight window in the demo app?

aelij:
@danwalmsley No black boxes on my machine.
For insight window, just type `(` :)

danwalmsley:
Ah I think id built the old version by mistake, looks great!

PR #16, opened 2017-05-01
Search panel

aelij:
Not working yet. When adding the panel to the `AdornerLayer`, its template isn't applied for some reason. Could someone please take a look?


danwalmsley:
I bet @grokys would know why straight away!
Has its style been loaded?

danwalmsley:
By style I mean template.xaml

danwalmsley:
I think you need to add an entry in the demo app.xaml for drop down.xaml and for searchpanel.xaml.

aelij:
I added them all to AvaloniaEdit.xaml. From what I can tell the XAML loads (OmniXAML complains about errors).

danwalmsley:
I should get chance to try it out tonight, will update you what I find.

PR #25, opened 2017-05-17
Minor fixes

danwalmsley:


src/AvaloniaEdit/Utils/TextFormatterFactory.cs
```diff
@@ -27,7 +27,7 @@ namespace AvaloniaEdit.Utils
     /// <summary>
     /// Creates TextFormatter instances that with the correct TextFormattingMode, if running on .NET 4.0.
     /// </summary>
-    internal static class TextFormatterFactory
+    public static class TextFormatterFactory
```
aelij: Why was this made public? It was internal in AvalonEdit. Besides, it doesn't do much, I was thinking of removing it altogether. You can just use `new TextFormatter()`.

src/AvaloniaEdit/Utils/ExtensionMethods.cs
```diff
@@ -27,7 +27,7 @@
 
 namespace AvaloniaEdit.Utils
 {
-    internal static class ExtensionMethods
+    public static class ExtensionMethods
```
aelij: I'm not sure I like the `Utils` namespace to be part of the API. Why do you need this?

PR #38, opened 2017-07-15
WIP: Added basic zoom scale with min and max properties.

danwalmsley:


danwalmsley:
Will close this as we require new apis in avalonia.

PR #48, opened 2018-03-13
Fix horizontal scroll

aelij:
I've left a few TODOs, but it's working


danwalmsley:
@aelij looks good, I found a few issues:

first one is this:
If I paste the following:
https://gist.githubusercontent.com/danwalmsley/6579ef522abb527402ae356ebd7914af/raw/41d2c05c4abddd5957ee5b5548059ca7e62fd9de/gistfile1.txt

I get this weird wrapping issue (ps this might not have been caused by the PR, but i didn't notice before)

image

See some lines take up the space of 2 lines, i looked but cant see why this is happening.

danwalmsley:
I also got it to go into an endless loop but need more time to find how to repro, only seen it in AvalonStudio so far.

danwalmsley:
Another issue is that if you scroll horizontally, it doesn't update the display until you do something like move the vertical scrollbar.

jeffreye:
Is there any other issues on this PR that I could work on?

danwalmsley:
Hey, yes if you try this branch then you will find scrolling doesnt seem to work correctly, like you cant scroll all the way to the bottom.

jeffreye:
> Hey, yes if you try this branch then you will find scrolling doesnt seem to work correctly, like you cant scroll all the way to the bottom.

Could you provide a sample? 
I remember that scroll size does not take nested controls into account, because it is only using number of lines of text and line height.

danwalmsley:
@jeffreye iv added a repro branch here: https://github.com/AvaloniaUI/AvaloniaEdit/tree/demo-scroll-issues

If used a grid splitter and you can see in the animation that as you resize the editor it doesn't update the scrolling viewport correctly so you are not able to scroll to the end...

ae-bug

PR #51, opened 2018-03-26
verify access checks against thread textdocument is created on.

danwalmsley:
Not neccesarily the ui thread.


danwalmsley:
This is how it was done in original avalon edit.

I came across the issue implementing Find In Files where you want to run on background thread and open each file as a TextDocument, then search through it.

Previously you couldnt because the only valid thread is ui thread, TextDocument isnt a UI component so it doesnt make sense, it only makes sense that it is owned an used from a single thread.

PR #56, opened 2018-05-31
fixes major bug in weak event manager where it would only unsubscribe

danwalmsley:
events that were of PropertyChangedHandlers meaning lots of events were being sent to the wrong instances.

### DIFF
--- a/src/AvaloniaEdit/AvaloniaEdit.csproj
+++ b/src/AvaloniaEdit/AvaloniaEdit.csproj
@@ -4,13 +4,7 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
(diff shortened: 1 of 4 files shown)


src/AvaloniaEdit/AvaloniaEdit.csproj
```diff
@@ -4,6 +4,14 @@
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <LangVersion>7.1</LangVersion>
```
aelij: I suggest you add this in a directory.build.props file (without the condition)

PR #59, opened 2018-08-03
Multithreading

jeffreye:
objects should be able to initialize in the background


danwalmsley:
Thanks 👍  I need to get the CI stuff fixed.

PR #70, opened 2018-11-25
Upgrade to Avalonia 0.7

aelij:
@danwalmsley Can you please publish to nuget.org? Thanks :)


KvanTTT:
Good! I wanted to create a new issue but you was faster :)

PR #71, opened 2019-01-07
Fix Search Panel

jeffreye:
* Fix no match popup
* Fix Bindings
* Fix text input

### DIFF
--- a/src/AvaloniaEdit/Search/SearchPanel.cs
+++ b/src/AvaloniaEdit/Search/SearchPanel.cs
@@ -21,6 +21,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Avalonia;
 using Avalonia.Controls;
+using Avalonia.Controls.Presenters;
 using Avalonia.Controls.Primitives;
 using Avalonia.Input;
 using Avalonia.Interactivity;
@@ -281,7 +282,7 @@ namespace AvaloniaEdit.Search
             base.OnTemplateApplied(e);
             _searchTextBox = e.NameScope.Find<TextBox>("PART_searchTextBox");
             _messageView = e.NameScope.Find<Popup>("PART_MessageView");
-            _messageViewContent = _messageView.Child as ContentControl;
+            _messageViewContent = _messageView.FindControl<ContentPresenter>("PART_MessageContent");
         }
 
         private void ValidateSearchText()
@@ -372,7 +373,7 @@ namespace AvaloniaEdit.Search
         }
 
         private Popup _messageView;
-        private ContentControl _messageViewContent;
+        private ContentPresenter _messageViewContent;
         private string _validationError;
 
         private void DoSearch(bool changeSelection)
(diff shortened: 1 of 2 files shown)


src/AvaloniaEdit/Search/SearchPanel.xaml
```diff
@@ -37,8 +37,17 @@
...
+                               StaysOpen="True"
+                               Focusable="False"
+                               IsOpen="False"
+                               ObeyScreenEdges="True">
+                            <ContentControl>Message Here</ContentControl>
+                            </Popup>
                         <ToggleButton Classes="ExpanderToggle" Name="Expander" VerticalAlignment="Top"
-                                      IsChecked="{TemplateBinding IsReplaceMode, Mode=TwoWay}"
+                                      IsChecked="{Binding IsReplaceMode, Mode=TwoWay, RelativeSource={RelativeSource TemplatedParent}}"
                                       Grid.Column="0"
```
aelij: If you just needed a two way binding, Avalonia has Mode in TemplateBinding

src/AvaloniaEdit/Search/SearchPanel.xaml
```diff
@@ -37,8 +37,17 @@
...
                     <Grid ColumnDefinitions="Auto,Auto,Auto"
                 RowDefinitions="Auto,Auto">
+                        <Popup Name="PART_MessageView"
+                               Classes="ToolTip"
+                               PlacementMode="Bottom"
+                               StaysOpen="True"
+                               Focusable="False"
+                               IsOpen="False"
+                               ObeyScreenEdges="True">
+                            <ContentControl>Message Here</ContentControl>
```
aelij: I think a `ContentPresenter` is good enough here. Also, add a named PART for the presenter, right now the code assumes a specific visual tree.

src/AvaloniaEdit/Search/SearchPanel.cs
```diff
@@ -456,8 +451,6 @@ private void SearchLayerKeyDown(object sender, KeyEventArgs e)
                     Close();
                     break;
             }
-
-            e.Handled = true;
```
aelij: Why this change?

jeffreye: In previous version, Search panel doesn't not take any input from keyboard

src/AvaloniaEdit/Search/SearchPanel.xaml
jeffreye: It was template binding previously, but it doesn't work.

src/AvaloniaEdit/Search/SearchPanel.xaml
jeffreye: Do your you mean use `ContentPresenter` instead of `ContentControl`?

aelij: Yes

aelij:
Thanks for the fix @jeffreye

PR #72, opened 2019-01-07
Open uri routed event

jeffreye:
OpenUri as a routed event. It should be handled by application (Open browser or navigate to another page in app). 

Similar to
https://docs.microsoft.com/en-us/dotnet/api/system.windows.documents.hyperlink.requestnavigate?view=netframework-4.7.2


aelij:
Thanks for the fix, @jeffreye

Issue #77, opened 2019-04-20
Update Avalonia to 0.8?


KvanTTT:
I updated Avalonia to 0.8 and now an app is crashing with the following exception :(

```
System.MissingMethodException: Method not found: 'Avalonia.Size Avalonia.Media.FormattedText.Measure()'.
  at at AvaloniaEdit.Text.TextLineRun.CreateRunForText(StringRange stringRange, TextRun textRun, Double widthLeft, Boolean emergencyWrap, Boolean breakOnTabs)
```

CreateLab:
+1 wait to solution

KvanTTT:
What about updated nuget? :pray:

danwalmsley:
@KvanTTY ping me a message on Monday on gitter and Il publish an updated nuget

Issue #79, opened 2019-05-11
Missing member exception


CreateLab:
install package from nuget
and use demo, when app start has this ecxeption
MissingMethodException: Method not found: 'Avalonia.Size Avalonia.Media.FormattedText.Measure()'

```
AvalonEdit:TextEditor  Name="Editor"
                                    Margin="30"
                                    SyntaxHighlighting="XML"
                                    HorizontalScrollBarVisibility="Auto"
                                    VerticalScrollBarVisibility="Visible"
                                    TextBlock.FontSize="30"/>
```

here all my code (little demo)
https://gitlab.com/maindlab/complab2/tree/master/CompLab2/GuiNote

PR #80, opened 2019-05-18
Update Avalonia to v0.8

gurgendav:
Fixes: #77, #79

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <LangVersion>7.1</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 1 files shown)


src/AvaloniaEdit.Demo/AvaloniaEdit.Demo.csproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Avalonia.Desktop" Version="$(AvaloniaVersion)" />
```
aelij: You should update `AvaloniaVersion` in `Directory.Build.props` instead

aelij:
Thanks for the fix

Issue #78, opened 2019-05-09
Fix copy/paste shortcuts on mac using ctrl instead of cmd


garyhertel:
The keyboard shortcuts on mac appear to be using the `ctrl` key instead of the more common `cmd`. While the windows ones still work on the mac, it does cause a bit of confusion since the Avalonia controls use `cmd` and the AvaloniaEdit control uses `ctrl`.

I asked about this in the chat:
```
Nikita Tsukanov@kekekeks Apr 09 22:55
Hotkeys are probably still set to Ctrl+C instead of Cmd+C
AvaloniaEdit needs to use hotkey info from AvaloniaLocator.Current.GetService<PlatformHotkeyConfiguration>();
```

garyhertel:
Thanks for fixing this! That will help clear up a lot of confusion on macs!

garyhertel:
This is still broken, it looks like this change reverted a lot of the fixes:
https://github.com/AvaloniaUI/AvaloniaEdit/commit/d578a9074d28b1113134e1a882ce1716c4b4d4cd#diff-35eccecfbec07325f9b15cae129fce3f

PR #81, opened 2019-05-18
Fix editor shortcuts on mac replacing ctrl with cmd

gurgendav:
* Replace shortcut for mac is set to CMD + ALT + F because CMD + H is used for "hide window" system command

Fixes: #78

### DIFF
--- a/src/AvaloniaEdit/Editing/CaretNavigationCommandHandler.cs
+++ b/src/AvaloniaEdit/Editing/CaretNavigationCommandHandler.cs
@@ -62,9 +62,14 @@ namespace AvaloniaEdit.Editing
         static readonly List<RoutedCommandBinding> CommandBindings = new List<RoutedCommandBinding>();
         static readonly List<KeyBinding> KeyBindings = new List<KeyBinding>();
 
-        private static void AddBinding(RoutedCommand command, KeyModifiers modifiers, Key key, EventHandler<ExecutedRoutedEventArgs> handler)
+        private static void AddBinding(RoutedCommand command, EventHandler<ExecutedRoutedEventArgs> handler)
         {
             CommandBindings.Add(new RoutedCommandBinding(command, handler));
+        }
+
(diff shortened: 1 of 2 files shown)


danwalmsley:
I think we added an api to avalonia that essentially does GetPlatformKey, I'll post a link to it.

gurgendav:
Okay. Please give me the link and I'll update my pull request

src/AvaloniaEdit/Editing/CaretNavigationCommandHandler.cs
```diff
@@ -108,7 +108,7 @@ static CaretNavigationCommandHandler()
             AddBinding(EditingCommands.MoveToDocumentEnd, InputModifiers.Control, Key.End, OnMoveCaret(CaretMovementType.DocumentEnd));
             AddBinding(EditingCommands.SelectToDocumentEnd, InputModifiers.Control | InputModifiers.Shift, Key.End, OnMoveCaretExtendSelection(CaretMovementType.DocumentEnd));
 
-            AddBinding(ApplicationCommands.SelectAll, InputModifiers.Control, Key.A, OnSelectAll);
+            AddBinding(ApplicationCommands.SelectAll, InputModifiers.None, Key.None, OnSelectAll);
```
aelij: Can you explain this change?

gurgendav: Sorry for really late response. 
It looked like some kind of workaround that key combinations for all ApplicationCommands was declared in class but for "SelectAll" doesn't. So I moved declaration to ApplicationCommands

aelij: Then I don't think this line is needed any more.

gurgendav: It's needed to add command binding. 

I created overload of AddBinding method which doesn't need those extra arguments.

aelij:
@gurgendav can you please merge with master? There are conflicts. Thanks

garyhertel:
It looks like a lot of this fix got reverted in the merge to mainline

PR #87, opened 2019-09-07
Fix searchpanel crash

HendrikMennen:
### What does this Pull request do?
- Update Avalonia to newest version
- Fix searchpanel crash
- Fix searchpanel cursor
- Edit replacenext.png

### How?
On the current avalonia version the searchpanel crashes because it can't find 
`_messageViewContent` 
using
 ` _messageView.FindControl<ContentPresenter>("PART_MessageContent");` 
I changed it to
`e.NameScope.Find<ContentPresenter>("PART_MessageContent");` 
which seems to work

I also changed the cursor used over the searchpanel control to be the default arrow instead of IBeam

Finally I edited the replace next icon which looked different than the other ones from:
searchbefore

to:
searchafter

**Yes, that was very important :)**


aelij:
Thanks for the fix :)

PR #88, opened 2019-09-08
Update + Multiple small bugfixes and improvements

HendrikMennen:
## What does this pull request do?
- Updated Avalonia to 0.8.999-cibuild0003807-beta
- Changed KeyGestures to work with newest avalonia
- Changed cursor on foldingmarker to default Arrow from IBeam
- `Caret.BringCaretToView(double border)` now uses the border parameter correctly
- The CompletionWindow now uses the parent window to position itself instead of using the MainWindow
- Fixed multiple ApplicationCommands in TextEditor so `TextEditor.CanUndo` , `TextEditor.CanRedo` , `TextEditor.Undo()`, `TextEditor.Redo()`, `TextEditor.Cut()`, `TextEditor.SelectAll()` and some more are working again 
- Since the ScrollTo functions are still not working I made the ScrollViewer public
- Fixed the style of Searchpanel to use theme colors and made the textboxes use labels if empty and not focused

Unbenannt

## How?
Obvious if you look at the code


src/AvaloniaEdit.Demo/AvaloniaEdit.Demo.csproj
```diff
@@ -21,7 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Avalonia.Desktop" Version="0.8.999-cibuild0003807-beta" />
+    <PackageReference Include="Avalonia.Desktop" Version="0.8.999-cibuild0004042-beta" />
```
aelij: Please use `$(AvaloniaVersion)` (see #90)

src/AvaloniaEdit/AvaloniaEdit.csproj
```diff
@@ -30,7 +30,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Avalonia" Version="0.8.999-cibuild0003807-beta" />
+    <PackageReference Include="Avalonia" Version="0.8.999-cibuild0004042-beta" />
```
aelij: Please use `$(AvaloniaVersion)` (see #90)

src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -115,7 +115,7 @@ protected override void OnTemplateApplied(TemplateAppliedEventArgs e)
             if (e.NameScope.Find("PART_CP") is ContentPresenter contentPresenter)
             {
                 contentPresenter.Content = TextView;
-                ((ISetLogicalParent)TextView).SetParent(this);
+                //((ISetLogicalParent)TextView).SetParent(this); Crash on newest avalonia build
```
aelij: I think it was needed for bindings to work. Check that it didn't break anything. Can you provide he stack trace?

src/AvaloniaEdit/Search/SearchPanel.xaml
```diff
@@ -174,4 +176,9 @@
             </ControlTemplate>
         </Setter>
     </Style>
+
+  <Style Selector="Button">
```
aelij: I think this style would apply to any button, you should constrain it to the above template

src/AvaloniaEdit/Search/TextBoxWithLabel.xaml.cs
```diff
@@ -0,0 +1,50 @@
...
+using Avalonia.Media;
+using Avalonia.Metadata;
+using Avalonia.Threading;
+using System;
+using System.ComponentModel;
+using System.Reflection;
+
+namespace AvaloniaEdit.Search
+{
+    public class TextBoxWithLabel : TextBox
```
aelij: Avalonia has this functionality built in to TextBox, just use the `Watermark` property.

src/AvaloniaEdit/TextEditor.cs
```diff
@@ -279,7 +279,7 @@ protected override void OnTemplateApplied(TemplateAppliedEventArgs e)
         /// Gets the scroll viewer used by the text editor.
         /// This property can return null if the template has not been applied / does not contain a scroll viewer.
         /// </summary>
-        internal ScrollViewer ScrollViewer { get; private set; }
+        public ScrollViewer ScrollViewer { get; private set; }
```
aelij: Why `public`?

aelij:
Thanks for the fix, @HendrikMennen! I will merge it after #81 since it has conflicting changes. You'll need to marge with master after that.

src/AvaloniaEdit.Demo/AvaloniaEdit.Demo.csproj
aelij: Perhaps also update to he latest version, same as other PR

src/AvaloniaEdit/Editing/TextArea.cs
HendrikMennen: My guess was that the new update sets the parent automaticly. I tested a lot and had no issues with just uncommenting it. The exception was that Parent is already set. Can send a full stack trace later

src/AvaloniaEdit/TextEditor.cs
HendrikMennen: "Since the ScrollTo functions are still not working I made the ScrollViewer public" I needed this to get an Event if the user scrolls

src/AvaloniaEdit/Search/SearchPanel.xaml
HendrikMennen: To any button that is in SearchPanel.xaml so I thought it was okay to do

HendrikMennen: Yeah fixed that

src/AvaloniaEdit/Editing/TextArea.cs
HendrikMennen: @danwalmsley did the same in #89

src/AvaloniaEdit/Search/TextBoxWithLabel.xaml.cs
HendrikMennen: Overlooked that. Now using Watermark instead

src/AvaloniaEdit/Editing/TextArea.cs
aelij: Then please remove this line

src/AvaloniaEdit/TextEditor.cs
aelij: I would rather you fixed the methods :) this should not be public.

aelij: You can also find the scrollviewer externally by searching the visual tree.

HendrikMennen: ok it is internal again

PR #90, opened 2019-09-13
Updated Avalonia

jp2masa:
## Changes
- Updated Avalonia.


aelij:
Thanks for the fix, but it's already being done in #88

aelij:
Actually, my mistake, this is a newer version. Merging :)

PR #92, opened 2019-10-21
Don't allow replace if editor is readonly

HendrikMennen:
Currently it is possible to use CTRL + H or the expander in the search field to open the replace option even if the editor is readonly. **This PR diables the replacemode if the TextEditor is in readonly**.

### DIFF
--- a/src/AvaloniaEdit/Search/SearchPanel.cs
+++ b/src/AvaloniaEdit/Search/SearchPanel.cs
@@ -107,7 +107,17 @@ namespace AvaloniaEdit.Search
         }
 
         public static readonly AvaloniaProperty<bool> IsReplaceModeProperty =
-            AvaloniaProperty.Register<SearchPanel, bool>(nameof(IsReplaceMode));
+            AvaloniaProperty.Register<SearchPanel, bool>(nameof(IsReplaceMode), validate: ValidateReplaceMode);
+
+        /// <summary>
+        /// Checks if replacemode is allowed
+        /// </summary>
+        /// <returns>False if editor is not null and readonly</returns>
+        private static bool ValidateReplaceMode(SearchPanel panel, bool v1)
+        {
+            if (panel._textEditor == null || !v1) return v1;
+            return !panel._textEditor.IsReadOnly;
+        }
 
         public bool IsReplaceMode
         {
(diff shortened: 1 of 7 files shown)


src/AvaloniaEdit/Search/SearchPanel.cs
```diff
@@ -190,29 +195,23 @@ private SearchPanel()
...
-                throw new ArgumentNullException(nameof(editor));
-            return Install(editor.TextArea);
-        }
+            var textArea = editor.TextArea;
+            if (textArea == null) throw new ArgumentNullException(nameof(textArea));
 
-        /// <summary>
-        /// Creates a SearchPanel and installs it to the TextArea.
-        /// </summary>
-        public static SearchPanel Install(TextArea textArea)
```
aelij: I would like to keep the ability to install directly into a `TextArea`, since it's a standalone control. Its also maintains API compact with the original project.

src/AvaloniaEdit/Search/SearchPanel.cs
```diff
@@ -190,29 +195,23 @@ private SearchPanel()
...
-        {
-            if (textArea == null)
-                throw new ArgumentNullException(nameof(textArea));
             var panel = new SearchPanel();
+
+            //Subscribe to readonly property
+            editor.GetObservable(TextEditor.IsReadOnlyProperty).Subscribe(x => panel.TextEditorReadOnly = x);
```
aelij: The property `TextEditorReadOnly` is not an `AvaloniaProperty` so it will not notify the binding When it's updated. I think a simpler approach could be to keep the `Install` methods as they were, add a `TextEditor` property, and bind to `TextEditor.IsReadOnly`.

aelij:
Thanks for the fix @HendrikMennen, please see my comments.

src/AvaloniaEdit/Search/SearchPanel.cs
```diff
@@ -111,7 +112,7 @@ public string SearchPattern
         public bool IsReplaceMode
         {
             get => GetValue(IsReplaceModeProperty);
-            set => SetValue(IsReplaceModeProperty, value);
+            set => SetValue(IsReplaceModeProperty, _textEditor?.IsReadOnly ?? false ? false : value);
```
aelij: AvaloniaProperties should not have additional code in their getters/setters since they won't necessarily be called (e.g. when using binding, animation, etc.) . You should place this in the value changed callback, set in `AvaloniaProperty.Register`.

HendrikMennen: I tried to do that and while the value stays correct the replace window still opens. I don't know why it is not working

src/AvaloniaEdit/Search/SearchPanel.cs
HendrikMennen: It is still there but it does not automatically install the search if you use TextArea as standalone anymore. Is this important?

PR #93, opened 2019-10-27
Fix line number font size 

HendrikMennen:
## Current behavior
If the fontsize is changed while the editor is visible the line numbers will not change their size
Unbenannt

## Expected behavior
Unbenannt

## How?
Just update the size before rendering using `EmSize = GetValue(TextBlock.FontSizeProperty);`
Maybe we should call the method `MeasureOverride(Size availableSize)` instead since it updates the fontsize aswell but I am not sure.

I added the ability to change the FontSize with CTRL+Scroll in the Demo.

### DIFF
--- a/src/AvaloniaEdit/Editing/LineNumberMargin.cs
+++ b/src/AvaloniaEdit/Editing/LineNumberMargin.cs
@@ -68,6 +68,8 @@ namespace AvaloniaEdit.Editing
         /// <inheritdoc/>
         public override void Render(DrawingContext drawingContext)
         {
+            EmSize = GetValue(TextBlock.FontSizeProperty);
+
             var textView = TextView;
             var renderSize = Bounds.Size;
             if (textView != null && textView.VisualLinesValid)
@@ -82,8 +84,7 @@ namespace AvaloniaEdit.Editing
                         Typeface, EmSize, foreground
                     );
                     var y = line.GetTextLineVisualYPosition(line.TextLines[0], VisualYPosition.TextTop);
-                    drawingContext.DrawText(foreground, new Point(renderSize.Width - text.Bounds.Width, y - textView.VerticalOffset),
-                        text);
+                    drawingContext.DrawText(foreground, new Point(renderSize.Width - text.Bounds.Width, y - textView.VerticalOffset), text);
                 }
             }
         }
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
@@ -6,6 +6,7 @@ using Avalonia;
 using Avalonia.Controls;
 using Avalonia.Controls.Platform;
 using Avalonia.Input;
+using Avalonia.Interactivity;
 using Avalonia.Markup.Xaml;
 using Avalonia.Media;
 using Avalonia.Media.Imaging;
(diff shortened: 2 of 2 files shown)


src/AvaloniaEdit/Editing/LineNumberMargin.cs
```diff
@@ -68,6 +68,8 @@ protected override Size MeasureOverride(Size availableSize)
         /// <inheritdoc/>
         public override void Render(DrawingContext drawingContext)
         {
+            EmSize = GetValue(TextBlock.FontSizeProperty);
```
aelij: Move inside `if` block below

src/AvaloniaEdit/Editing/LineNumberMargin.cs
```diff
@@ -82,8 +84,7 @@ public override void Render(DrawingContext drawingContext)
                         Typeface, EmSize, foreground
                     );
                     var y = line.GetTextLineVisualYPosition(line.TextLines[0], VisualYPosition.TextTop);
-                    drawingContext.DrawText(foreground, new Point(renderSize.Width - text.Bounds.Width, y - textView.VerticalOffset),
```
aelij: Please avoid unnecessary formatting changes as they make PR review harder

src/AvaloniaEdit/Editing/LineNumberMargin.cs
HendrikMennen: I believe I found a better solution. Now it updates the fontsize and the margin correctly

aelij:
@HendrikMennen I've gone back to check how the original AvalonEdit dealt with this, and it seems like they register to the `TextView.VisualLinesChanged` event and use `InvalidateVisual()`:

https://github.com/icsharpcode/AvalonEdit/blob/8c8ca8844151f3b6677cb3234652a9aa984ecd4f/ICSharpCode.AvalonEdit/Editing/LineNumberMargin.cs#L96-L111

(note you don't need to assign `TexaArea` as it's handled by the base class)

https://github.com/icsharpcode/AvalonEdit/blob/8c8ca8844151f3b6677cb3234652a9aa984ecd4f/ICSharpCode.AvalonEdit/Editing/LineNumberMargin.cs#L162-L165

So I'd appreciate it if you could do the same here.

src/AvaloniaEdit/Editing/LineNumberMargin.cs
```diff
@@ -108,6 +123,13 @@ private void OnDocumentLineCountChanged(object sender, EventArgs e)
             OnDocumentLineCountChanged();
         }
 
+        void TextViewVisualLinesChanged(object sender, EventArgs e)
+        {
+            InvalidateVisual();
+            InvalidateMeasure();
```
HendrikMennen: `InvalidateVisual()` allone was not enough to make it work.

aelij: What about invalidate measure by itself?

src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -106,6 +106,8 @@ protected TextArea(TextView textView)
             //{
             //    TextView.SetScrollOffset(new Vector(_offset.X, _offset.Y * TextView.DefaultLineHeight));
             //});
+
+            //Set's correct font size to linenumber
```
aelij: Remove comment

src/AvaloniaEdit/Editing/LineNumberMargin.cs
HendrikMennen: for this exact scenario with font size it works

aelij:
Thanks, @HendrikMennen !

PR #116, opened 2020-10-06
Take TSQL Highlighting Syntaxe from AvalonEdit.

ricardoromaobr:
Add a TSQL Highlighting  from AvalonEdit. With this is possible to create SQL Editor for SQL Server TSQL editor.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/AvaloniaUI/AvaloniaEdit?pullRequest=116) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/AvaloniaUI/AvaloniaEdit?pullRequest=116) before we can accept your contribution.<br/><br/>:x: ricardoromaobr [sign now](https://cla.dotnetfoundation.org/AvaloniaUI/AvaloniaEdit?pullRequest=116)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/AvaloniaUI/AvaloniaEdit?pullRequest=116) it.</sub>

src/AvaloniaEdit/AvaloniaEdit.csproj
```diff
@@ -33,7 +33,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Avalonia" Version="$(AvaloniaVersion)" />
+    <PackageReference Include="Avalonia" Version="0.10.0-preview5" />
```
danwalmsley, line 36: should be defined here:
https://github.com/AvaloniaUI/AvaloniaEdit/blob/master/Directory.Build.props#L5

danwalmsley (changes requested):
Looks good... just a minor change required.

src/AvaloniaEdit/Highlighting/Resources/Resources.cs
```diff
@@ -60,6 +60,7 @@ internal static void RegisterBuiltInHighlightings(HighlightingManager.DefaultHig
 			                         "XML-Mode.xshd");
             hlm.RegisterHighlighting("XAML", new[] { ".xaml", ".paml", ".axaml" }, "XML-Mode.xshd");
 			hlm.RegisterHighlighting("MarkDown", new[] { ".md" }, "MarkDown-Mode.xshd");
+			hlm.RegisterHighlighting("TSQL", new[] { ".sql" }, "TSQL-Mode.xshd");
```
KvanTTT: Not only T-SQL files have `.sql` file extension but PL/SQL, MySql, SQLite, and other dialects.

ricardoromaobr: I take it exactly it is in the AvalonEdit. What I see in the demo that the editor is setting by the name not by the extension with the property **SyntaxHighlighting**, so I believe  that who is writing a tools need to take care about it. Identified by the the database provider some other things to know what type of the database need to be used.

KvanTTT: > I take it exactly it is in the AvalonEdit.

It doesn't mean that it's true. I suggest renaming it just to `SQL`.

ricardoromaobr: What I'm saying is that in first place people connect to database and after that work with it. Every tools that I works todays the query editor is created by the existing connection.If somebody want to write a offline query, a tool the he is using need to be a way to say what the target database to be used to setting the editor. There is not another way to do it.

Other thing that would be made is a  SyntaxHighting for Ansi-SQL and when the database is not knowledge setting the ansi-sql

ricardoromaobr: You are thinking  in put all others dialects in the same file SQL-Mode.xshd?

KvanTTT: > What I'm saying is that in first place people connect to database and after that work with it.

Code analysers does not use database connection, at least not necessarily.

I mean SQL is a more common naming compare to TSQL.

ricardoromaobr: But if I need an editor just for PL/SQL and I make a mistake and use for example GETDATE() (T-SQL) that is not valid key word on PL/SQL. And I want to mark this word as  invalid or not knowledge. I have not experience with AvalonEdit parse. So I will do what you say to do. But I want to learn how it works.

KvanTTT: Ok, maybe you are right and there is no need to unify dialects since the extension is not used for dialect detection.

ricardoromaobr: No. You are correct. To do good things need to be careful and from my home to my jobs I take one hours driven and I come think  about it. The Highlighting is provide just by scan using the resulting tokens (in the case of AvaloniaEdit by  Regular Expression), and the correct  sintaxe by the parse that I think that the AvaloniaEdit  doesn't made it. It need to be implemented.

Issue #109, opened 2020-08-24
Is there a way to underline text like in AvalonEdit?


RomanShumikhin:
Is there a way to underline text like in AvalonEdit?
```
...TextRunProperties.SetTextDecorations(TextDecorations.Underline);
```
I see that `TextDecorations.Underline` exists, but I am not sure how to use it.
For context: I am writing a custom colorizer by implementing DocumentColorizingTransformer.
Thank you!

PR #135, opened 2021-05-23
support underline and strikethrough

whistyun:
## Related issues
#109

## Details of changes
* add new property strikethrough.
* draw underline and strikethrough if it is requred.
* increase a text element height for underline spacing.

## Screenshots
image


Gillibald:
It seems there wasn't any progress in the past so I opened up a new PR with your changes and slightly adjusted the underline and strikethrough measurements.

Closing this in favor of the new one.

PR #145, opened 2021-07-26
Fix background color in template

BAndysc:
In the TextEditor control template, background wasn't set, therefore you couldn't set control background using Background property.

This change matches original AvalonEdit: https://github.com/icsharpcode/AvalonEdit/blob/master/ICSharpCode.AvalonEdit/TextEditor.xaml#L14


maxkatz6 (approved):
Thanks

PR #148, opened 2021-08-05
Fix scroll after resize with 125% DPI

### LINKED ISSUE #147: Scroll bug after resize with DPI set to 125%
Here is the repro: https://recordit.co/OHc6Jj7uqu
To reproduce-> Set DPI on Windows to 125%->Paste a lot of code to the AvalonEdit (i used TextView file) ->resize window->navigate down the text field with your arrows->hold "down" arrow.
This PR probably caused this issue - #139
Adding such code fixed issue for me but probably I would need to investigate more to understand why it happens
image

Takoooooo:
Our code doesn't match upstream https://github.com/icsharpcode/AvalonEdit/blob/master/ICSharpCode.AvalonEdit/Rendering/TextView.cs#L1185
Also fixes that problem on 175% DPI
closes #147

### DIFF
--- a/src/AvaloniaEdit/Rendering/TextView.cs
+++ b/src/AvaloniaEdit/Rendering/TextView.cs
@@ -1215,7 +1215,7 @@ namespace AvaloniaEdit.Rendering
             }
 
             // Apply final view port and offset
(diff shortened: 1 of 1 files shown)


danipen:
@Takoooooo verified that it works for me.

@danwalmsley could we move it to be merged as soon as possible? thanks guys!

PR #156, opened 2021-11-03
Fixed search results appearing under the search panel

### LINKED ISSUE #155: Search results appear under the search panel
Some search results appear under the search panel and it's difficult to find them.

danipen:
Closes #155.

When calling BringCaretToView, take into account the height of the search panel and give that margin (plus an extra line height) to ensure that the search result is visible.


danipen (commented):
I verified that when you select text in the editor and you show the search panel, it performs a search before the Template was applied, so the _border is null. LGTM.

danwalmsley (approved):

PR #158, opened 2021-11-10
Fix selection issues

Takoooooo:
If I will click the right mouse button on a single word it will be selected, I think it is not expected behaviour and it doesn't match AvalonEdit behaviour.
If I will click the right mouse button on the selected text selection will be re-set. All those issues are caused because we wire to the event that will send both left and right mouse button click, while we expect only left.
Here is the gif demonstrating the issue.
Peek 2021-11-10 10-36


danipen:
@Takoooooo I tested your changes, and they work ok. 

However, I found a couple of extra issues:

- Triple-click doesn't work (it should select the whole line).
- The right-click pointer down should move the caret to the clicked position.

Could you take a look? Thanks!

PR #165, opened 2021-11-27
Fixed empty lines height

danipen:
This PR makes empty lines have the same height normal lines have.

### Problem explanation
AvaloniaEdit assigned different heights for empty lines and nonempty lines.
image

The reason is that when creating TextLines with no runs (empty lines), the **extra +2 pixels for underlining was not being added**.

### How I fixed it
Refactored how LineHeight and BaseLine values are calculated to a commonplace, and use the same source code when calculating the height for empty and for non-empty lines.

### DIFF
--- a/src/AvaloniaEdit/Text/TextLineRun.cs
+++ b/src/AvaloniaEdit/Text/TextLineRun.cs
@@ -65,11 +65,9 @@ namespace AvaloniaEdit.Text
 
         public static double GetDefaultLineHeight(FontMetrics fontMetrics)
         {
-            // add additional space for underline and line gap
-            return fontMetrics.LineHeight +
-                fontMetrics.UnderlinePosition +
-                fontMetrics.UnderlineThickness + 
-                fontMetrics.LineGap;
+            // adding an extra 15% of the line height look good across different font sizes
+            double extraLineHeight = fontMetrics.LineHeight * 0.15;
+            return fontMetrics.LineHeight + extraLineHeight;
         }
 
         public static double GetDefaultBaseline(FontMetrics fontMetrics)
--- a/src/AvaloniaEdit/Text/TextLineRun.cs
+++ b/src/AvaloniaEdit/Text/TextLineRun.cs
@@ -1,14 +1,13 @@
 using System;
 using Avalonia;
 using Avalonia.Media;
+using Avalonia.Media.TextFormatting;
 
 namespace AvaloniaEdit.Text
 {
     internal sealed class TextLineRun
     {
         private const string NewlineString = "\r\n";
-        private const double BaselineFactor = 0.1;
-        private const double HeightFactor = 1.2;
 
         private FormattedText _formattedText;
         private Size _formattedTextSize;
@@ -41,7 +40,8 @@ namespace AvaloniaEdit.Text
                     var box = embeddedObject.ComputeBoundingBox();
                     return box.Y;
                 }
-                return GetDefaultBaseline(FontSize);
+
+                return GetDefaultBaseline(TextRun.Properties.FontMetrics);
             }
         }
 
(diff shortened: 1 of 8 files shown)


src/AvaloniaEdit/Text/TextLineRun.cs
```diff
@@ -58,11 +58,22 @@ public double Height
...
 
+        public static double GetDefaultLineHeight(double fontSize)
+        {
+            // add 2pt for underline
+            return fontSize * HeightFactor + 2;
+        }
+
+        public static double GetDefaultBaseline(double fontSize)
+        {
+            return fontSize * BaselineFactor;
```
Gillibald, line 74: We should use the FontMetrics here instead of using hard codes numbers.

https://github.com/AvaloniaUI/Avalonia/blob/7842883961d094e08e9def7f30cf32fd573179c7/src/Avalonia.Visuals/Media/TextFormatting/FontMetrics.cs

danipen:
@Gillibald please let me know if my changes using FontMetrics looks good now. Thank you!

src/AvaloniaEdit/Text/TextLineRun.cs
```diff
@@ -59,19 +59,22 @@ public double Height
...
             }
         }
 
-        public static double GetDefaultLineHeight(double fontSize)
+        public static double GetDefaultLineHeight(FontMetrics fontMetrics)
         {
-            // add 2pt for underline
-            return fontSize * HeightFactor + 2;
+            // add additional space for underline and line gap
+            return fontMetrics.LineHeight +
```
Gillibald, line 69: LineHeight already includes the LineGap. We don't need to add the UnderlinePosition to the line height because the position is always within the LineHeight.

danipen: @Gillibald if I don't add extra height the lines are very close. Let me show you some screenshots...

danipen: Using just `fontMetrics.LineHeight`, makes lines to be too close compared with VSCode, for example.
image

What should I do, use an extra hardcoded margin?

danipen: Adding those extra `1.740234375` px added by adding `fontMetrics.UnderlinePosition + fontMetrics.UnderlineThickness` makes AvaloniaEdit to look exactly like VSCode.

danipen: Should I calculate that needed `1.740234375` extra height as a percentage of the line height?

danipen: Something like this?
```
        public static double GetDefaultLineHeight(FontMetrics fontMetrics)
        {
            double factor = 1.15;
            return fontMetrics.LineHeight * factor;
        }
```

Gillibald: This line is wrong https://github.com/AvaloniaUI/Avalonia/blob/master/src/Skia/Avalonia.Skia/FormattedTextImpl.cs#L557 so you need to add the LineGap on your own.

danipen: @Gillibald, Any recommended calculation for the line gap?

danipen: Something like this?
```
        public static double GetDefaultLineHeight(FontMetrics fontMetrics)
        {
            // FontMetrics doesn't include a lineGap to the total lineHeight
            // so we need to add the LineGap on our own
            double lineGap = fontMetrics.LineHeight * 0.15;
            return fontMetrics.LineHeight + lineGap;
        }
```

Gillibald: Hmm then there must be something else wrong / different. A LineGap of zero is totally fine. As long as the underline renders just fine I don't see any issue. In general you can't compare the rendering results of two different text rendering stacks. There might be different rounding / scaling in place etc.

Gillibald: Also it should be fine to add some extra space  to each line as you have done. This should rely on some fixed factor and the current fontSize.

danipen: Adding an extra 15% of the line height works fine for different font sizes. I'll change that. Thank you!

Gillibald (approved):
LGTM

PR #166, opened 2021-12-01
Added an exception handler to the TextMate installation

danipen:
This PR adds a new method in `AvaloniaEdit.TextMate.TextMate` class that allows registering an exception handler. 
```
        public static void RegisterExceptionHandler(Action<Exception> handler)
        {
            _exceptionHandler = handler;
        }
```
The exception handler is invoked whenever an exception is thrown in any AvaloniaEdit event handler.


gwynroberts (commented):
Look good

gwynroberts (approved):

PR #169, opened 2021-12-14
You should be able to move caret by right clicking text area

Takoooooo:
Originally it was requested by @danipen 
This feature is not present in the original AvaloniaEdit but it would not be bad if we add that IMO. It is what 90% of text editors do.

### DIFF
--- a/src/AvaloniaEdit/Editing/SelectionMouseHandler.cs
+++ b/src/AvaloniaEdit/Editing/SelectionMouseHandler.cs
@@ -16,13 +16,13 @@
 // OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
+using Avalonia;
+using Avalonia.Input;
+using AvaloniaEdit.Document;
+using AvaloniaEdit.Utils;
 using System;
 using System.ComponentModel;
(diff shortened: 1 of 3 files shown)


src/AvaloniaEdit/Editing/SelectionMouseHandler.cs
```diff
@@ -392,8 +392,15 @@ private void TextArea_OptionChanged(object sender, PropertyChangedEventArgs e)
...
         private void TextArea_MouseLeftButtonDown(object sender, PointerPressedEventArgs e)
         {
-            if (e.GetCurrentPoint(TextArea).Properties.IsLeftButtonPressed)
-            { 
+            if (e.GetCurrentPoint(TextArea).Properties.IsLeftButtonPressed == false)
+            {
+                if (TextArea.IsRightClickMovesCaret == true)
+                {
+                    SetCaretOffsetToMousePosition(e);
+                }
```
Takoooooo, line 400: should we ignore that if PointerPressedEventArgs is handled already?

grokys: Probably doesn't make much difference in this case, but can't hurt.

danipen (approved):
LGTM, the naming changes proposed by @grokys sounds good.

danipen (approved):

PR #170, opened 2021-12-17
Fix triple click

Takoooooo:
Triple-click should select the whole line but before this PR it wasn't working.
**Note**: this code is ported from AvalonEdit


danipen (approved):
LGTM

PR #183, opened 2022-01-12
[NEEDS HELP] Refactor TextMateSharp integration

Takoooooo:
This PR refactors TextMateSharp integration. Previously we weren't able to publish AvaloniaEdit and AvaloniaEdit.TextMate to nuget because AvaloniaEdit.TextMate contained ~5mb of grammars and themes that would massively bloat our dll size and also AvaloniaEdit was suitable to work only on Desktop, it wasn't compatible with WASM at all because it worked only with the filesystem.
What I did to refactor it:

- Introduced [IResourceStorage](https://github.com/Takoooooo/AvaloniaEdit/blob/63e4c0bd601d2ebda58f5c8281cc1d020cfda3d6/src/AvaloniaEdit.TextMate.Shared/Storage/Abstractions/IResourceStorage.cs) that is needed to prevent TextMate from accessing filesystem and also sometimes TextMate may access the system even when you don't expect it which maybe not a good idea.
- Needed to do a [PR](https://github.com/danipen/TextMateSharp/pull/8) to a TextMateSharp to rework the way it works with resources.
- Added helper methods that helps users to fill their own ResourceStorages.
- Made a cleanup of TextMate-related code.
- Minified all big .json files so dll with resources is ~4.5mb instead of ~5.3mb.

How to set up TextMate with new API:
```csharp
 new Installation(_textEditor, ResourceLoader.SetupStorage(ThemeName.DarkPlus, GrammarName.CSharp))
```
Code above will use VS Dark theme by default and CSharp grammar and syntax highlighting will just work for C#.

**PROBLEMS:**

1. Setting default grammar doesn't work *somewhy*,in IGrammarStorage I have a concept like SelectedGrammar, it should allow users to set the default theme. If I use the code below for setting default grammar the result is random, sometimes it works, sometimes some parts of code arent highlighted.
image
The result looks like that:
image

2. Some grammars don't work currently. For example, switching to C++ will throw an exception, it tries to locate grammar by a scopename source.asm which I *have no idea where to find*.

**MERGE USING THIS OPTION**
image


danipen:
I'm trying to debug what is wrong with your implementation. I found this. It seems that the root rule is not correctly parsed. I'm trying to figure out if the reason for this issue is the refactor or minifying the grammar files.
 
image

danipen:
Minifying the grammar files is not the reason. Trying to debug the issue. I'm unable to focus in what could be wrong because there are a lot of changes in the refactor.

danipen:
Sorry, I'm getting crazy with this. 

After the refractor, the way grammars are loaded sometimes causes the grammar to be corrupted. It seems to be a concurrency issue, but it's being hard to debug. The worst part is that I cannot dedicate much time to this, because I have other obligations to attend to.

I think that the smartest approach right now is just separate assemblies and leave the code as it was since it worked fine.

**I'll start a new PR to extract the grammars to `AvaloniaEdit.TextMate.Grammars`.**

Let's integrate that into master so we unlock the wasm part and then let's think if a refactor is needed.

danipen:
@Takoooooo I created a new PR that extracts all the grammars and the themes to a new assembly. I extensively tested it, and it seems to work like expected (I found no issues):

https://github.com/AvaloniaUI/AvaloniaEdit/pull/185

PR #186, opened 2022-01-14
Improve TextMate performance

danipen:
**This PR makes TextMate 450% faster**

### Rework the TextEditorModel

Before this PR, the line contents must be obtained using `Dispatcher.UIThread.InvokeAsync`, due to a restriction in AvaloniaEdit, that requires the LineDocuments accessed from the UI thread.

I reworked the [TextEditorModel](https://github.com/AvaloniaUI/AvaloniaEdit/blob/master/src/AvaloniaEdit.TextMate/TextEditorModel.cs) to use [AvaloniaEdit's document snapshots](https://github.com/AvaloniaUI/AvaloniaEdit/blob/5856b95ce53b8f509f770e5718031983508dc162/src/AvaloniaEdit/Document/TextDocument.cs#L345). Document snapshots are thread-safe, so we can avoid calling `Dispatcher.UIThread.InvokeAsync` every time we need to get a line content or the line length. I added unit tests for that.

### Rework the TextMateColoringTransformer

I'm also reworked the [TextMateColoringTransformer](https://github.com/AvaloniaUI/AvaloniaEdit/blob/master/src/AvaloniaEdit.TextMate/TextMateColoringTransformer.cs) to:
- track the first and last visible line in the document.
- then only set tokens for visible lines, instead of for the whole document.
- just redraw the lines that have been retokenized, to avoid transforming lines that did not change the tokenization state.

Those changes together make AvaloniaEdit TextMate integration much faster when handling mid/big files (I detected it was not fast enough for mid/big files) and it reduces CPU use a lot.

### Demo
This is AvaloniaEdit managing a really large C# files **(133.442 lines)** (the file is a[ big class](https://github.com/AvaloniaUI/Avalonia/blob/master/src/Avalonia.Controls/Grid.cs), copied and pasted x40 times):

https://user-images.githubusercontent.com/501613/149822699-c19dbbf6-5471-4a02-944e-cf25969156be.mp4

### DIFF
--- a/src/AvaloniaEdit.TextMate/TextEditorModel.cs
+++ b/src/AvaloniaEdit.TextMate/TextEditorModel.cs
@@ -165,7 +165,7 @@ namespace AvaloniaEdit.TextMate
                     UpdateLine(startLine);
                 }
 
-                if (e.RemovalLength > 0 || e.InsertionLength > 0)
+                if (e.RemovalLength - e.InsertionLength != 0)
                 {
                     UpdateLineRanges();
                 }
--- a/src/AvaloniaEdit.TextMate/TextEditorModel.cs
+++ b/src/AvaloniaEdit.TextMate/TextEditorModel.cs
@@ -152,9 +152,9 @@ namespace AvaloniaEdit.TextMate
 
                 UpdateSnapshot();
 
-                // invalidate current and previous line
+                // invalidate the changed line it's previous line
                 // some grammars (JSON, csharp, ...)
-                // need to invalidate previous lines
+                // need to invalidate the previous line
                 if (startLine - 1 >= 0)
                     InvalidateLine(startLine - 1);
                 InvalidateLine(startLine);
--- a/src/AvaloniaEdit.TextMate/TextEditorModel.cs
+++ b/src/AvaloniaEdit.TextMate/TextEditorModel.cs
@@ -11,10 +11,8 @@ namespace AvaloniaEdit.TextMate
 {
     public class TextEditorModel : AbstractLineList
     {
-        private object _lock = new object();
         private readonly TextDocument _document;
         private readonly TextView _textView;
-        private volatile int _lineCount;
         private DocumentSnapshot _documentSnapshot;
         private Action<Exception> _exceptionHandler;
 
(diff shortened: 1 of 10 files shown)


src/AvaloniaEdit.TextMate/TextEditorModel.cs
```diff
@@ -66,33 +69,19 @@ private void TextView_ScrollOffsetChanged(object sender, EventArgs e)
...
-                if (e.RemovedText is { })
+                if (e.RemovalLength > 0)
                 {
                     var startLine = _document.GetLineByOffset(e.Offset).LineNumber - 1;
                     var endLine = _document.GetLineByOffset(e.Offset + e.RemovalLength).LineNumber - 1;
 
                     for (int i = endLine; i > startLine; i--)
                     {
                         RemoveLine(i);
+                        _documentSnapshot.LineCount--;
```
danwalmsley, line 84: Why doesnt `_documentSnapshot.Update(e)` do this?

danwalmsley (commented):
just a question...

src/AvaloniaEdit.TextMate/TextEditorModel.cs
danipen: This happens before the document is actually modified.

danipen: Long answer:

We need to remove lines from the model in the `DocumentChanging` event (they cannot be processed in the `DocumentChanged` event because we would be accessing deleted offsets).

The `_documentSnapshot.Update(e)` is processed in. the `DocumentChanged`, so unfortunately we need to separate that logic.

So summarizing:
- We must process removed lines in the  `DocumentChanging` event.
- In the time-lapse between we remove lines to the model, and we actually update the line ranges, we need to actually say the `TMModel` we have less lines, until the `DocumentSnapshot` is fully updated.
- This is the reason we need to decrement the LineCount. Otherwise the `TMModel` would be accessing indexes out of range in the tokenizer thread.

danwalmsley: ok

danwalmsley (approved):

PR #210, opened 2022-02-15
Use textmatesharp grammars nuget

danipen:
Use [TextMateSharp.Grammars](https://www.nuget.org/packages/TextMateSharp.Grammars/) nuget instead of maintaining here the project `Avalonia.TextMate.Grammars`.

That sources have been moved here: https://github.com/danipen/TextMateSharp/tree/master/src/TextMateSharp.Grammars

### DIFF
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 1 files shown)


src/AvaloniaEdit.Demo/MainWindow.xaml.cs
```diff
@@ -17,7 +17,8 @@
 using AvaloniaEdit.Folding;
 using AvaloniaEdit.Rendering;
 using AvaloniaEdit.TextMate;
-using AvaloniaEdit.TextMate.Grammars;
+
```
Takoooooo, line 20: nit: extra line

PR #213, opened 2022-02-22
Try/catch exceptions for override methods and events

danipen:
It prevents the app from crashing if something goes wrong.


miryamGSM (approved):
LGTM

Issue #194, opened 2022-01-27
Wrong complete window position on Avalonia 0.10.11 and 0.10.12


alexeybut:
Complete window opens on a wrong position, it AvaloniaEdit built on Avalonia 0.10.11 or 0.10.12.
With 0.10.10 the position is correct.
Platform: MS Windows 10.
The attached picture illustrates the issue:

![popup-wrong-place](https://user-images.githubusercontent.com/8170797/151433087-f03bbd62-98dc-43ef-a2f5-5d6a1abcfc99.png)
.

cordasfilip:
@danipen Is there a way to fix this with the current release?

danipen:
@whistyun any ideas about how to fix in current release?

cordasfilip:
@whistyun I even tried using preview
<PackageReference Include="Avalonia" Version="11.0.0-preview2" />
<PackageReference Include="Avalonia.AvaloniaEdit" Version="11.0.0-preview1" />
<PackageReference Include="Avalonia.Desktop" Version="11.0.0-preview2" /> 
not working

whistyun:
I think AvaloniaEdit shows three types of windows: CompletionWindow, OverloadInsightWindow and IME window.
@cordasfilip Did you mean the IME window has a wrong position?

Only the position of the IME window is wrong in my environment.

**CompletionWindow:**  
![image](https://user-images.githubusercontent.com/13712028/201506115-cd46dde7-773c-4598-bce5-896707a93e74.png)  

**OverloadInsightWindow:**  
![image](https://user-images.githubusercontent.com/13712028/201506154-bfb29c5f-862b-43f1-bbf8-9ff77efd16f2.png)

**IME window(I turned on IME, and inputed "世界"):**  
![image](https://user-images.githubusercontent.com/13712028/201506780-759f9db6-3fa3-4165-ae76-e0f1f50f1b2e.png)

I try it on [a test project](https://github.com/whistyun/testcode-garage/tree/master/TestForAvaloniaEdit).

cordasfilip:
@whistyun Strange it's working fine on the project you sent me thanks for help will investigate if any other library is I am using could be causing the issue.

danipen:
@Gillibald do you think the IME issue is already fixed or needs a fix?

whistyun:
@danipen I forget to compare preview1 and latest.
[...]

PR #215, opened 2022-03-08
fix the wrong complete window position

whistyun:
about https://github.com/AvaloniaUI/AvaloniaEdit/issues/194.

In Avalonia 0.10.11, Avalonia [observes the position of the popup parent](https://github.com/AvaloniaUI/Avalonia/blob/089fda0dc7fedd67ed6a4febd2be8c843760eb55/src/Avalonia.Controls/Primitives/Popup.cs#L430), and moves popup accordingly.

Currently, `CompletionWindowBase` dosen't change `PlacementTarget`. This causes that the position of the popup parent is treated as a position of the popup ([L380](https://github.com/AvaloniaUI/Avalonia/blob/089fda0dc7fedd67ed6a4febd2be8c843760eb55/src/Avalonia.Controls/Primitives/Popup.cs#L380) and [L388](https://github.com/AvaloniaUI/Avalonia/blob/089fda0dc7fedd67ed6a4febd2be8c843760eb55/src/Avalonia.Controls/Primitives/Popup.cs#L388)).
So [Host.ComfigurePosition](https://github.com/whistyun/AvaloniaEdit/blob/7a7c8990db326b0a89611d777eb7b6e84403101c/src/AvaloniaEdit/CodeCompletion/CompletionWindowBase.cs#L375) is treated as changing the position of the popup parent.
As a result, The position of the popup is re-moved to `(HorizontalOffset`, `VerticalOffset`).

This PR fixes the problem by changing properties affect the position of the popup.

### DIFF
--- a/src/AvaloniaEdit/CodeCompletion/CompletionWindowBase.cs
+++ b/src/AvaloniaEdit/CodeCompletion/CompletionWindowBase.cs
@@ -76,15 +76,20 @@ namespace AvaloniaEdit.CodeCompletion
         /// Creates a new CompletionWindowBase.
         /// </summary>
         public CompletionWindowBase(TextArea textArea) : base()
-        {     
+        {
             TextArea = textArea ?? throw new ArgumentNullException(nameof(textArea));
             _parentWindow = textArea.GetVisualRoot() as Window;
 
-           
+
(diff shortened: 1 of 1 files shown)


danipen (approved):

danipen:
Fixes #194

PR #216, opened 2022-03-09
[WIP] TextFormatter port

Gillibald:


danipen:
I gave it a try to it. Good progress! 

I found some issues so far, that you probably noticed:
- The intra-line spacing is smaller now. Can we mimic old metrics?
image
- Empty lines are overlapped:
image
- The syntax highlight is broken.
- The caret only works fine for the first line
- Editing doesn't work fine. It offsets the caret while typing (I only tried editing the first line).

danipen:
- [ ] After pasting an emoji, such as ❤, the text font on the right is changed:
font-changed

danipen:
- [ ] `IndexOutOfRangeException` opening a binary file as text.
Yes, I know it's an extreme case, but I think it's worth protecting this kind of exceptions. Reproduced:
image

Gillibald:
I wonder how TabGlypRun ever worked. The run itself should reuse measurements of the replaced sequence otherwise the general text layout changes when `ShowTabs` is enabled.

danipen:
Although the implementation was a mess, it was fixed in [this PR](https://github.com/AvaloniaUI/AvaloniaEdit/pull/192) and it worked fine :-)

https://user-images.githubusercontent.com/501613/151114591-829ec35a-6186-4a77-b66c-144a380cdc00.mp4

Gillibald:
I had an idea about long lines. What if we enforce line wrapping if we encounter a certain line length? We might get vertical virtualization that way and it is easier to edit in theory.

HendrikMennen:
I think the default should be to just cut the lines off after a specific length and show a few dots, like VSCode default behaviour:
image

Of course it would be nice to have both options.

danipen:
Well. I'm pretty sure forcing the user to wrap long lines will end in a bunch of support requests for us 😂.

We already did that in the past for the macOS editor, and some users handling minified javascript hit that issue and complained a lot (they don't care about long lines but they don't want to polish the document, wrapping long lines).

Using the VSCode approach could be an option. The best editor for handling long lines is Sublime Text. It completely virtualizes lines horizontally. Not sure how, but it does. It just displays the whole line, and you can navigate/edit/select text without any delay.

BTW the current version of AvaloniaEdit is able to do that. For long lines, it just assumes the same len for every glyph (when using a monospaced font), and it splits line runs in runs of MAX 3000 chars. Can't we do the same with the new implementation?

Gillibald:
Avalonia's TextFormatter is still not 100% matching WPF's version so we need to slightly adjust text processing at certain points. In general this port is mostly working and I am happy that RTL support is also kinda working so we are making huge progress here.

src/AvaloniaEdit/Rendering/VisualLineTextParagraphProperties.cs
```diff
@@ -0,0 +1,46 @@
...
+		internal TextWrapping textWrapping;
+		internal double tabSize;
+		internal double indent;
+		internal bool firstLineInParagraph;
+
+		public override double DefaultIncrementalTab => tabSize;
+
+		public override FlowDirection FlowDirection => FlowDirection.LeftToRight;
+		public override TextAlignment TextAlignment => TextAlignment.Left;
+		public override double LineHeight => DefaultTextRunProperties.FontRenderingEmSize * 1.2;
```
Gillibald, line 37: @danipen This is just some number I brought up maybe you can find some factor that produces similar line spacing to the old version

danipen: The LineHeight is not being applied to the first line:
image

danipen:
@Gillibald I found some more issues. I detail them here for reference:

- [ ] the First line doesn't respect the specified line height.
image

- [ ] Using option `ShowTabs` is not fine yet: the caret isn't well-positioned, and the mouse hit test is wrong:
view-tabs

- [ ] View spaces option doesn't work as expected:
view-spaces

- [ ] View EOL option makes lines higher.
view-eol

src/AvaloniaEdit/Rendering/CollapsedLineSection.cs
```diff
@@ -27,82 +27,76 @@ namespace AvaloniaEdit.Rendering
...
+#if DEBUG
 		internal string Id;
-	    private static int _nextId;
-		#else
-		internal const string Id = "";
-		#endif
-		
+		private static int _nextId;
+#else
+		const string ID = "";
```
hez2010, line 36:
```suggestion
		const string Id = "";
```

Otherwise, it will break compilation in Release config.

Gillibald:
Fist we need to create a stable branch for the current 0.10.x release
If nothing major comes up we might be able to merge this week.

mattj23:
I assume you guys already know this, but just in case not it seems that (at least on the latest version) word wrap removes the spacing for blank lines:

image


image

mattj23:
I suspect this is also related, but once word wrap is on the line spacing for different sized fonts gets lost.

Without word wrap:

image

With word wrap:

image

As always, I'm happy to do anything I can to help.  This is a great project and, despite some of the setbacks, the performance improvements I've seen on this branch vs the main branch are really impressive.

Gillibald:
Thanks for testing this helps a lot 👍
Will have a deeper look at this scenario shortly

Gillibald:
@mattj23 Could you test my latest changes? TextWrapping should now work with variable size lines.

mattj23:
@Gillibald, for some reason I can't get the nightly build you used, I'm getting this error:

```  
AvaloniaEdit.csproj: [NU1102] Unable to find package Avalonia with version (>= 0.10.999-cibuild0020141-beta)
  - Found 197 version(s) in Avalonia Nightly [ Nearest version: 0.10.999-cibuild0020138-beta ]
  - Found 63 version(s) in nuget.org [ Nearest version: 0.10.13 ]
```

I'll give it a try on a different machine tomorrow and let you know!

mattj23:
@Gillibald, I tested the latest commit and as best I can tell you've fixed those issues I saw with the blank line height and the variable text height. Looks great.

danipen:
- [ ] Failing tests:
- `Visual_Line_Should_Create_Two_Text_Lines_When_Wrapping`
- `Visual_Line_Should_Create_One_Text_Lines_When_Not_Wrapping`

danipen:
- [ ] selection decorator is not accurate:
weird-selection

danipen:
- [ ] `InvalidOperationException` undoing editions (Ctrl + Z).
exception-undoing-changes

Gillibald:
I have fixed some hit testing and rendering issues here: https://github.com/AvaloniaUI/Avalonia/pull/8094/files

Will test my changes against AvaloniaEdit and will fix the remaining issues

workgroupengineering:
Hi @Gillibald, do you have any idea when this PR might be merge?

Gillibald:
This is only blocked by some compiler issues in the test project. Once that is sorted out we can merge.

PR #218, opened 2022-03-18
Expose the search panel in the TextEditor

danipen:
Expose the `SearchPanel` in the `TextEditor` so users can, manage the search panel programmatically (for example, show, hide, change the search term, etc ...).

Additionally, fixed an issue with the search panel control template. When the search panel was visible, and then you detached the `TextEditor` from the logical tree, and you attached it later (i.e moving the control from one panel to another), the SearchPanel's control template was lost. I fixed it by closing the SearchPanel when detaching from the logical tree and restoring it if needed when attaching.


miryamGSM (approved):
LGTM

PR #260, opened 2022-09-29
Implement ITextInputMethodClient for TextArea

Gillibald:


src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -745,12 +756,14 @@ protected override void OnGotFocus(GotFocusEventArgs e)
             // First activate IME, then show caret
```
danipen, line 756: Does the order matters? Old comments said:
First activate IME, then show caret

src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -745,12 +756,14 @@ protected override void OnGotFocus(GotFocusEventArgs e)
...
             //ime.OnGotKeyboardFocus(e);
```
danipen, line 757: We can delete old commented code.

src/AvaloniaEdit/Editing/TextArea.cs
```diff
@@ -745,12 +756,14 @@ protected override void OnGotFocus(GotFocusEventArgs e)
...
             Caret.Show();
+            _imClient.SetTextArea(this);
         }
 
         protected override void OnLostFocus(RoutedEventArgs e)
         {
             base.OnLostFocus(e);
             Caret.Hide();
+            _imClient.SetTextArea(null);
             //ime.OnLostKeyboardFocus(e);
```
danipen, line 767: We can delete old commented code.

danipen (changes requested):

src/AvaloniaEdit/Editing/TextArea.cs
Gillibald: This is handled by Avalonia so a control that wants to interact with IME just has to implement `ITextInputMethodClient`

Gillibald: Will change order just for consistency

danipen (approved):

PR #263, opened 2022-10-06
Fixes VisualLine leak

Gillibald:
This PR introduces a fix that removes a VisualLineDrawingVisual from the logical tree on VisualLine.Dispose

### DIFF
--- a/src/AvaloniaEdit/Rendering/TextView.cs
+++ b/src/AvaloniaEdit/Rendering/TextView.cs
@@ -1045,8 +1045,8 @@ namespace AvaloniaEdit.Rendering
         private VisualLine BuildVisualLine(DocumentLine documentLine,
                                    TextRunProperties globalTextRunProperties,
                                    VisualLineTextParagraphProperties paragraphProperties,
-                                   VisualLineElementGenerator[] elementGeneratorsArray,
-                                   IVisualLineTransformer[] lineTransformersArray,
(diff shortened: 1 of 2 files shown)


src/AvaloniaEdit/Rendering/TextView.cs
```diff
@@ -890,7 +890,7 @@ protected override Size MeasureOverride(Size availableSize)
             {
                 // no document -> create empty list of lines
                 _allVisualLines = new List<VisualLine>();
-                _visibleVisualLines = new ReadOnlyCollection<VisualLine>(_allVisualLines.ToArray());
+                _visibleVisualLines = new ReadOnlyCollection<VisualLine>(_allVisualLines);
```
danipen, line 901: @Gillibald be careful with this change. I think we need to maintain generating a new collection. I already fixed that in [this commit](https://github.com/AvaloniaUI/AvaloniaEdit/commit/d5c80250404a576115df2c82a414102ba2796cdd).

See upstream project implementation:
- https://github.com/icsharpcode/AvalonEdit/blob/b6fe189e7be4a8b650c0c670fdac8b839800d9cc/ICSharpCode.AvalonEdit/Rendering/TextView.cs#L927
- https://github.com/icsharpcode/AvalonEdit/blob/b6fe189e7be4a8b650c0c670fdac8b839800d9cc/ICSharpCode.AvalonEdit/Rendering/TextView.cs#L1025

danipen (changes requested):

Issue #272, opened 2022-10-13
Folding crash


HendrikMennen:
To reproduce, simply open the xml example and put `<!--` in front of one of the nodes, then collapse it.
![image](https://user-images.githubusercontent.com/25281882/195565064-e2ade9c7-f8d2-4783-b378-c0ff4cf266c9.png)
```
Unhandled exception. System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at Avalonia.Media.TextFormatting.Unicode.Codepoint.ReadAt(ReadOnlySlice`1 text, Int32 index, Int32& count) in /_/src/Avalonia.Base/Media/TextFormatting/Unicode/Codepoint.cs:line 170
   at Avalonia.Media.TextFormatting.Unicode.LineBreakEnumerator.NextCharClass() in /_/src/Avalonia.Base/Media/TextFormatting/Unicode/LineBreakEnumerator.cs:line 295
   at Avalonia.Media.TextFormatting.Unicode.LineBreakEnumerator.MoveNext() in /_/src/Avalonia.Base/Media/TextFormatting/Unicode/LineBreakEnumerator.cs:line 57
   at Avalonia.Media.TextFormatting.TextFormatterImpl.TryGetLineBreak(TextRun textRun, LineBreak& lineBreak) in /_/src/Avalonia.Base/Media/TextFormatting/TextFormatterImpl.cs:line 454
   at Avalonia.Media.TextFormatting.TextFormatterImpl.FetchTextRuns(ITextSource textSource, Int32 firstTextSourceIndex, TextEndOfLine& endOfLine, Int32& textSourceLength) in /_/src/Avalonia.Base/Media/TextFormatting/TextFormatterImpl.cs:line 407
   at Avalonia.Media.TextFormatting.TextFormatterImpl.FormatLine(ITextSource textSource, Int32 firstTextSourceIndex, Double paragraphWidth, TextParagraphProperties paragraphProperties, TextLineBreak previousLineBreak) in /_/src/Avalonia.Base/Media/TextFormatting/TextFormatterImpl.cs:line 62
[...]

PR #273, opened 2022-10-14
Attempt to prevent folding crashes

Gillibald:
Fixes: #272

### DIFF
--- a/src/AvaloniaEdit/Rendering/VisualLineText.cs
+++ b/src/AvaloniaEdit/Rendering/VisualLineText.cs
@@ -69,7 +69,14 @@ namespace AvaloniaEdit.Rendering
 				offset,
 				DocumentLength - relativeOffset);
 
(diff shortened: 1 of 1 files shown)


HendrikMennen:
For me this fixes the issue :)

Numpsy:
The error I was seeing seems ok with this change as well :-)

danipen:
@HendrikMennen can you describe the issue? I disabled `AllowToScrollBelowDocument` and it seems to work fine for me:
folding

HendrikMennen:
If I collapse something at the bottom of an element it looks like this.
image

Once I force it to redraw by resizing or scrolling the issue is resolved.

danipen:
Could you provide a repro case in the demo application?

HendrikMennen:
Okay nevermind it was my fault, sorry! It works perfectly now :)

danipen:
Glad to hear that!

PR #284, opened 2022-10-25
Fixed lines not be highlighted sometimes

danipen:
Fixed a condition that caused that some lines were not highlighted until they were re-created.

- Open a document of, let's say 500 lines and go to line 400 (visible range is 400-450).
- Open a document of 100 lines (first line is visible).
- When the `TMModel` starts to emit tokens, the visual lines haven't been constructed yet.

So `_firstVisibleLineIndex` and `_lastVisibleLineIndex` are not updated and still 400 and 450. So this condition avoids not post to the dispatcher the redraw

```csharp
    bool changedLinesAreNotVisible =
        (firstChangedLineIndex < _firstVisibleLineIndex && lastChangedLineIndex < _firstVisibleLineIndex) ||
        (firstChangedLineIndex > _lastVisibleLineIndex && lastChangedLineIndex > _lastVisibleLineIndex);

        if (changedLinesAreNotVisible)
            return;
```

**Solution:**
Just initialize control a flag `_areVisualLinesValid` and check it to calculate if `changedLinesAreNotVisible`. 
Otherwise, post all the redraw events to the dispatcher to ensure that lines are redrawn after transforming them.

### DIFF
--- a/src/AvaloniaEdit.TextMate/TextMateColoringTransformer.cs
+++ b/src/AvaloniaEdit.TextMate/TextMateColoringTransformer.cs
@@ -27,6 +27,7 @@ namespace AvaloniaEdit.TextMate
         private TextView _textView;
         private Action<Exception> _exceptionHandler;
 
+        private volatile bool _areVisualLinesValid = false;
         private volatile int _firstVisibleLineIndex = -1;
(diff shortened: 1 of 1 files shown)


miryamGSM (approved):
LGTM

PR #318, opened 2023-03-22
Upgrade to Avalonia 11.0.0-preview6

aldelaro5:
I am leaving this here because I tried to upgrade it and while everything seemed fine and builds, I am hitting a very bad performance issue which profiling seems to suggests that the GC is put under a lot of pressure which didn't happen with preview5. The CPU usage is much higher than it should and the app becomes very unresponsive.

Just in case I can't get anywhere so my branch is there.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <AvaloniaVersion>11.0.999-cibuild0031007-beta</AvaloniaVersion>
+    <AvaloniaVersion>11.0.0-preview6</AvaloniaVersion>
     <TextMateSharpVersion>1.0.50</TextMateSharpVersion>
     <NewtonsoftJsonVersion>13.0.1</NewtonsoftJsonVersion>
     <VersionSuffix>beta</VersionSuffix>
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
@@ -49,7 +49,7 @@ namespace AvaloniaEdit.Demo
             _textEditor.ShowLineNumbers = true;
             _textEditor.ContextMenu = new ContextMenu
             {
-                Items = new List<MenuItem>
+                ItemsSource = new List<MenuItem>
                 {
                     new MenuItem { Header = "Copy", InputGesture = new KeyGesture(Key.C, KeyModifiers.Control) },
                     new MenuItem { Header = "Paste", InputGesture = new KeyGesture(Key.V, KeyModifiers.Control) },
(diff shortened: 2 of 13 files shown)


danipen:
Could you please share your profiling data? What are the hot paths?

danipen:
How do you reproduce the bad performance? Do you experience it in your application or just in the AvaloniaEdit demo project?

aldelaro5:
> Could you please share your profiling data? What are the hot paths?

This is how I got there, I am unfortunately not experienced with profiling so I actually have a hard time figuring out where are the hot paths. Rider doesn't really show me something that jumps out to me. What I do know is the GC has a lot more pressure,
preview6:
Screenshot from 2023-03-22 12-04-25

preview5:
Screenshot from 2023-03-22 12-05-43


> How do you reproduce the bad performance? Do you experience it in your application or just in the AvaloniaEdit demo project?

The demo project. I just needed to run it to see the issues, it's incredibly unresponsive immediately and all the time.

danipen:
Quite good perf under windows for me ... Are you trying windows OS?
perf

danipen:
@aldelaro5 do you experience the same bad perf if you execute the application directly without the debugger?

aldelaro5:
I am on arch linux, I have access to a windows vm I can try that later, but in digging more, I might have a hint after analysing more profiling:

Screenshot from 2023-03-22 12-53-12

This path is notably hot in preview 6, but not in preview 5. I also noticed the RunLoop has a lot of "lock contention" which sounds like this hot path could be stalling the UI thread?

It does seem to involve skia, maybe it's a linux specific issue?

Gillibald:
Is MatchFamily called more often than TryCreateGlyphTypeface?

aldelaro5:
Ok it seems to be an avalonia issue.

This pr https://github.com/AvaloniaUI/Avalonia/pull/10455 is the one that seems to have introduced this problem because it seems if I try ci builds before this pr, the perf comes back, but if I try the ci build of this pr, it reproduces.

aldelaro5:
> Is MatchFamily called more often than TryCreateGlyphTypeface?

I think it's more saying that either MatchFamily is slow OR that it ends up being called using this path so much that this path as a whole is slow. Given the pr I posted above seems the most suspicious, it's most likely the later it seems?

Numpsy:
fwiw, I've given it a try in both a Windows desktop app and in web assembly, and it seems to work ok, with perf about the same as the preview 2 release

aldelaro5:
I just filed an avalonia issue about this because I noticed the same degradation in a completely different context. I can however confirm it doesn't seem to affect Windows so this seems to be a Linux specific issue.

Numpsy:
So, is this perf issue a blocker to getting a preview 6 compatible AvaloniaEdit build released, or as the issue is in Avalonia can we still get a new build here whilst waiting for that? (I just think a new compatible build would be nice for testing other things is all(

Numpsy:
Thanks

bouyeijiang:
The Linux mint system I am using also has the same issue, the window system is fine

PR #320, opened 2023-04-02
Update TextMateSharp and drop Newtonsoft.Json

hez2010:
- Update TextMateSharp: the new version uses System.Text.Json instead of Newtonsoft.Json, and also brings trimming+AOT support


Numpsy:
Would it be possible to get a new preview release build with the package updates and avalonia 11 preview 6 changes in it? (I've updated TextMateSharp locally in my app and tested that works, but need to use a local AvaloniaEdit build to get the preview6 changes as it stands)

danipen:
@Takoooooo

PR #321, opened 2023-04-02
Enable compiled bindings

hez2010:
Make completion list trimming compatible.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -3,8 +3,7 @@
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AvaloniaVersion>11.0.0-preview6</AvaloniaVersion>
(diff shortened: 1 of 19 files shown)


danipen:
Building on top of this comment. 
https://github.com/AvaloniaUI/Avalonia/pull/10880#issuecomment-1493372348

Should we avoid enabling by default here?

Numpsy:
I've been trying to get an app updated from Avalonia 11 preview 4 and the related lib versions to the latest versions of things, with compiled bindings and trimming enabled, and AvaloniaEdit appears to be working there (though i'm not using completions).

Are compiled bindings supposed to be disabled at the top of CompletionList.xaml ?

hez2010:
I tested the demo app, and the CompletionList works fine.

danipen:
cc/ @Gillibald @danwalmsley @Takoooooo @kekekeks @grokys

Gillibald (approved):

PR #335, opened 2023-06-01
Update to Avalonia 11 RC

Gillibald:


src/AvaloniaEdit/RoutedCommand.cs
```diff
@@ -84,6 +86,23 @@ void ICommand.Execute(object parameter)
             add { }
             remove { }
         }
+
+        IInputElement GetGloballyFocusedElement()
```
maxkatz6, line 90: Would be more reliable to handle global focus events

maxkatz6 (approved):

PR #341, opened 2023-06-08
Change ICompletionData.Image from Bitmap to IImage

mgarstenauer:
I propose changing `ICompletionData.Image` from `Bitmap` to `IImage`. Having `Bitmap` in the interface is unnecessarily restrictive. Changing the type to `IImage` allows other types of image sources. For example: `CroppedBitmap`, `DrawingImage`

In my case, I have an icon font that has a custom `IImage` implementation.

⚠ Note that this is a breaking change! Any application implementing `IComplectionData` (like AvaloniaEdit.Demo) will have to be updated. However, I think that this change is worthwhile and should be made before releasing 11.0.0.


danipen (approved):
LGTM.

PR #343, opened 2023-06-09
Fix: Column ruler behind current line highlight

mgarstenauer:
**Problem:**
The vertical column ruler is rendered behind the current line highlight.

**Fix:**
I suggest changing `CurrentLineHighlightRenderer.Layer` to `KnownLayer.Background`. This way, the vertical column ruler and other highlights (such as search result markers) can be rendered on top.

In `TextView` the `ColumnRulerRenderer` needs to be registered after the `CurrentLineHighlightRenderer`.

Also fixed a typo in the existing code.


danipen (approved):
LGTM.

PR #345, opened 2023-06-09
Fix: CompletionList style is broken.

mgarstenauer:
**Problem:**
The default `CompletionList` style is broken. It currently works in `AvaloniaEdit.Demo` because the style is overridden in `App.xaml`. When removing the style in `App.xaml`, it looks like this:
image

The tooltip position is also off.

**Fix:**
I removed the style from `App.xaml` and fixed the default style in `CompletionList.xaml`.
I also fixed the tooltip positioning in `CompletionWindow.cs`.

Tested with with the following themes: Fluent/Light, Fluent/Dark, Simple/Light, Simple/Dark

image

### DIFF
--- a/src/AvaloniaEdit/CodeCompletion/CompletionList.xaml
+++ b/src/AvaloniaEdit/CodeCompletion/CompletionList.xaml
@@ -1,13 +1,11 @@
 <Styles xmlns="https://github.com/avaloniaui"
-        
         xmlns:cc="clr-namespace:AvaloniaEdit.CodeCompletion;assembly=AvaloniaEdit"
-        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        x:CompileBindings="False">
+        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
 
     <Style Selector="cc|CompletionList">
         <Setter Property="Template">
             <ControlTemplate>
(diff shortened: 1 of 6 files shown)


src/AvaloniaEdit/CodeCompletion/CompletionList.xaml
```diff
@@ -1,13 +1,11 @@
 <Styles xmlns="https://github.com/avaloniaui"
-        
         xmlns:cc="clr-namespace:AvaloniaEdit.CodeCompletion;assembly=AvaloniaEdit"
-        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        x:CompileBindings="False">
```
mgarstenauer, line 5: Note that `x:CompileBindings="False"` is the reason why the completion list is empty. I am actually not sure why.

danipen (approved):
Looks good.

src/AvaloniaEdit/CodeCompletion/CompletionList.xaml
danipen: Any ideas about this @Gillibald @maxkatz6 ?

maxkatz6: Not sure why either, but I assume some CLR members were trimmed as unused, since they were only referenced with non-compiled (reflection) bindings.

PR #346, opened 2023-06-10
Port styles to control themes

mgarstenauer:
This PR ports the styles to control themes (following the [Porting to Control Themes](https://github.com/AvaloniaUI/Avalonia/wiki/Porting-to-Control-Themes) guide).

Changes:
- The styles are unchanged, all controls still look the same. (I didn't fix anything. The `InsightWindow`, for example, is not yet theme-aware, only works for "Light" themes. This PR doesn't fix that.)
- Formatting/indentation of the existing XAML files was mixed. I tried to normalize the format, following a style close to AvaloniaUI.
- Documented `CollapseIfSingleOverloadConverter` and made it public because it is used in the styles. It should be public in case someone wants to copy/override the style.
- Removed `DropDownButton`. This is not used anywhere. I recommend using the official AvaloniaUI `DropDownButton` should the need arise.
- Fixed a typo.

Tests:
- I tested `AvaloniaEdit.Demo` with all theme variants: Fluent/Light, Fluent/Dark, Simple/Light, Simple/Dark
- `<Design.PreviewWith>` for the `SearchPanel` also works in the Designer (JetBrains Rider).

I'll deal with any merge conflicts as they come up.


maxkatz6 (approved):
Thanks!
Looks good to me.
I will do some minor changes making internal XAML files invisible outside separately.

maxkatz6:
Also need to fix dark/light theme switching

PR #354, opened 2023-07-15
Styling adjustments

maxkatz6:
Continuing #194

Changes:
1. Cleanup: replaced "clr-assembly:" with "using:"
2. Instead of adding a style on every ContentControl, introduced CompletionTipContentControl type specifically for auto-completion tooltip content.
3. Removed VisualLineDrawingVisual style as it seems to do nothing. Cursor property is inheritable and in the demo app cursor on inline controls works as expected. @danipen please let me know if I am wrong here.
4. Made internal xaml files actually internal, so they can't be accessed directly with ResourceInclude, and so they can be merged together into a single dictionary
5. Introduced OverloadViewer*** and TextAreaSelectionBrush brushes to fix insight window theming
6. Restructured Base, Fluent and Simple files, so Base one can be used directly without referencing any specific theme. User just need to define brushes as they expect.


maxkatz6:
@danwalmsley @Gillibald @grokys

jmacato (approved):
lgtm

Gillibald (approved):

PR #358, opened 2023-07-18
Allow customisation of the keys that trigger insertation of a completion

zenonet:
Some people might not want to insert code completions using the enter key because they want to be able to create line breaks even when a completion window shows up. For example they might only want to insert completions using the tab key.

This PR adds a `Key[]` called `CompletionAcceptKeys` to the `CompletionList` Class. Now, when a keypress is handled by the completion list, it will check if the pressed key is is present in the CompletionAcceptKeys array instead of deciding based on the hard coded switch statement.

By default, this array will contain the Enter key and the tab key, so the behavior is unchanged.


zenonet:
@dotnet-policy-service agree

src/AvaloniaEdit/CodeCompletion/CompletionList.cs
```diff
@@ -163,13 +174,14 @@ public void HandleKey(KeyEventArgs e)
...
                     break;
-                case Key.Tab:
-                case Key.Enter:
-                    if(CurrentList.Count > 0)
-                    {
-                        e.Handled = true;
-                        RequestInsertion(e);
-                    }                    
+                default:
+                    if (CompletionAcceptKeys.Contains(e.Key))
```
danipen, line 178: Can you group the two conditions in one using && to avoid 2-level nesting?

danipen (changes requested):
LGTM. Thanks for the contribution. Can you please apply the requested changes just for code readability?

zenonet:
Done!

danipen (approved):
Looks good.

PR #371, opened 2023-09-06
Migrate to Avalonia 11.0.4

danipen:
No changes. Just use the new version.

@maxkatz6 please note that the `WeakReferenceTests.cs` were not migrated to the headless platform. I changed to migrate them, but they are not passing anymore in the headless platform. Please, take a look if you want to re-enable them.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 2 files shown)


Gillibald (approved):

danipen:
@Takoooooo are you planning to generate nugets?

Takoooooo:
@danipen I guess we agreed with team members to not migrate NuGet's to minor 11.0.x versions to not force people to update their apps  to minor versions

danipen:
Ok. But note that without these changes the build is broken. Cc @maxkatz6

PR #372, opened 2023-09-21
Fix potential dead lock in within SetClipboardText

Gillibald:


grokys (approved):

src/AvaloniaEdit/Editing/EditingCommandHandler.cs
```diff
@@ -426,7 +426,7 @@ private static void SetClipboardText(string text, Visual visual)
         {
             try
             {
-                TopLevel.GetTopLevel(visual)?.Clipboard?.SetTextAsync(text).GetAwaiter().GetResult();
+                TopLevel.GetTopLevel(visual)?.Clipboard?.SetTextAsync(text);
```
alexhost, line 429: @Gillibald `SetTextAsync` is asynchronous and exceptions will not be caught. Perhaps you have exception handling further out, but if so the try-catch is pointless. Any reason why `SetClipboardText` cannot be async?

PR #373, opened 2023-09-26
Fix ForegroundTextTransformation objects mem leak

danipen:
To avoid heap mem pressure, we're using an [ArrayPool](https://github.com/AvaloniaUI/AvaloniaEdit/blob/2a07c6ea07ddce749b85b902ffff02aa7f7e1276/src/AvaloniaEdit.TextMate/TextMateColoringTransformer.cs#L131) to avoid too many `ForegroundTextTransformation` allocations. 

The issue is that if this class holds a reference to this (`TextMateColoringTransformer`) via `ForegroundTextTransformation.IColorMap` interface implementation. This causes all `ForegroundTextTransformations ` that live in memory (in the shared array pool) will be referencing the `TextMateColoringTransformer ` instance, driving a mem leak. 

The issue can be avoided by passing the dictionary with the color map to the `ForegroundTextTransformation` instead of passing an interface.


danipen:
@Gillibald please could you review these changes?

Gillibald (approved):

PR #386, opened 2023-12-08
[WIP] Fix Completion Tooltip Placement

HendrikMennen:
## EDIT
I noticed in some cases with long completionlists this does not work well, I will try to find another way.

When scrolling down the tooltip position cannot always be taken from the container, because the container is not always generated at this point. I fixed this by simply placing the tooltip to the bottom whenever this happens.

## Current Behaviour:
image

## Expected:
image

### DIFF
--- a/src/AvaloniaEdit/CodeCompletion/CompletionWindow.cs
+++ b/src/AvaloniaEdit/CodeCompletion/CompletionWindow.cs
@@ -61,7 +61,6 @@ namespace AvaloniaEdit.CodeCompletion
             {
                 IsLightDismissEnabled = true,
                 PlacementTarget = this,
-                Placement = PlacementMode.RightEdgeAlignedTop,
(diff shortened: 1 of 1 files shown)


maxkatz6 (approved):
Feels like it new or old code shouldn't be necessary, if we set tooltip on the specific ListBoxItem with right placement mode. But either way, a new code looks like a reasonable workaround.

HendrikMennen:
> Feels like it new or old code shouldn't be necessary, if we set tooltip on the specific ListBoxItem with right placement mode. But either way, a new code looks like a reasonable workaround.

I already tried setting the tooltip to the ListBoxItem directly, and used it this way for some time in my own fork, but this caused another issue:
When the ListBoxItem is very long, that it enables horizontal scrolling, the tooltip will float far offscreen. 
image
The only way I know to fix this would be with calculating an offset, any different idea?

PR #393, opened 2024-01-09
Fix replace next functionality

### LINKED ISSUE #355: "Replace next" incorrect behavior
### The "replace next" function of the search panel implements an incorrect behavior when used more than once.

- Version:

AvaloniaEdit 11.0.0

- Simptoms:

If you have a text that contains multiple matches and replace one by one using the "Replace next" button it skips one of the matches each time.

- Cause:

Internally the "Replace next" function calls "FindNext", which searches one char ahead of the caret position, this behavior makes sense for the "FindNext" functionality but not for the replace as once you replace a match it moves the caret to the next one so the next time you press the "Replace next" button that match is skipped and the next one is replaced.

- Solution:

"Replace next" should use its own search criteria that searches from the caret position, not from the next position.

- Considerations:

"FindNext" also may present an erratic behavior because of the search-ahead, for example, if an user places the caret at the beginning of a file and the searched text is present starting at the first char the function will skip it and go to the next. The search-ahead behavior should be used only after the first "FindNext" is issued and if the user has not moved manually the caret.

danipen:
If you had a text that contains multiple matches and replace one by one using the "Replace next" button it skips one of the matches each time (see #355).

Changes on this PR:

* Re-enable existing (but commented) `FindTests`.
* Added unit tests to cover this functionality.
* Fixes #355.

### DIFF
--- a/src/AvaloniaEdit/Search/SearchPanel.cs
+++ b/src/AvaloniaEdit/Search/SearchPanel.cs
@@ -371,12 +371,19 @@ namespace AvaloniaEdit.Search
                 foreach (var result in _strategy.FindAll(_textArea.Document, 0, _textArea.Document.TextLength).Cast<SearchResult>())
                 {
                     _renderer.CurrentResults.Add(result);
-                    if (changeSelection && result.StartOffset >= offset)
-                    {
+                }
+
+                if (changeSelection)
+                {
+                    // select the first result after the caret position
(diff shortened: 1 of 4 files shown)


Gillibald (approved):

danipen:
@Gillibald there are new changes since your review just in case you want to take a look. Also feel free to merge it's already tested.

PR #399, opened 2024-02-16
Fix undo (Ctrl+Z) exception after not successful pasting

### LINKED ISSUE #398: If pasting fails, Undo throws Exception
If we try to paste something other than text - like image from snipping tool. 

1. Pasting fails **1** text wil be null and OnPaste exits. 
2. problem is that **2** begin edit group was allready opened. After this "pasting" und operation (Ctrl +Z by default) will throw.. 

image

 (_undoGroupDepth becomes 1, 

image

I checked AvalonEdit source but implementation is very different (and problem do not occurs). 
**I will oper PR with sollution.**

KrzysztofDusko:
fixes #398

### DIFF
--- a/src/AvaloniaEdit/Editing/EditingCommandHandler.cs
+++ b/src/AvaloniaEdit/Editing/EditingCommandHandler.cs
@@ -506,7 +506,10 @@ namespace AvaloniaEdit.Editing
                 }
 
                 if (text == null)
+                {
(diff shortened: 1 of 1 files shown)


KrzysztofDusko:
test fails but.. Its not my fault ..
@danipen, @maxkatz6 its very simple PR can you merge this into master ? 
image

maxkatz6 (approved):

PR #402, opened 2024-03-07
LineNumbersMargin property added to TextEditor

NickRimmer:
Hey, community (;

I want to suggest a tiny update, to be able to set margins for line numbers.


Numpsy:
I can't really comment on the usefulness of the API, but maybe it could have the same change as in #405 to use ```GetBindingObservable``` instead of ```new Binding``` ?

NickRimmer:
> I can't really comment on the usefulness of the API, but maybe it could have the same change as in #405 to use `GetBindingObservable` instead of `new Binding` ?

Absolutely agree. Will push updated soon. Thank you (;

NickRimmer:
My changes rebased to latest AvaloniaEdit master with suggested fixes

maxkatz6 (approved):

PR #406, opened 2024-03-19
[Breaking] Use CommandManager from Labs

maxkatz6:
Avalonia.Labs now has CommandManager with RoutedCommand partially ported from this repository and WPF with some extra features.
Old implementation in this repository can be replaced by Labs package.
https://github.com/AvaloniaUI/Avalonia.Labs/releases/tag/11.0.10.1

Note: it's a huge binary breaking change for anybody, who referenced routed command, and source breaking change for these who used routed command. 

If merged, then only when we ready to do such changes.


maxkatz6:
This PR should wait for 12.0 or at least 11.1.

heartacker:
？

PR #415, opened 2024-04-19
Enable overstrike mode

mgarstenauer:
This PR enables *Overstrike* mode (in other editors called *OVR, Overwrite, Overtype*), which is not yet working.

Simply uncommenting the code in *TextArea.cs* solves the problem. All relevant code paths for Overstrike from AvalonEdit are already present in AvaloniaEdit.

Instruction for testing:
- Run *AvaloniaEdit.Demo*.
- Press `Insert` (`ins`) key on the keyboard.

Expected behavior:
- The caret changes from a line to block.
- Typing now overwrites the right characters instead of inserting characters at the caret position.


danipen (approved):
LGTM

PR #419, opened 2024-04-27
Remove margin around TextEditor

mgarstenauer:
This is a cosmetic PR that removes the 30px margin around the `TextEditor` in the demo.

Before:
image

After:
image

(Looking at the git history, the `Margin` has been there for the last 7+ years. I don't think that the margin serves any purpose. Let's remove it to avoid any confusion!)


Rekkonnect:
Most code editors have a margin on the line number panel to make it more readable and prettier. Removing the margin feels much worse to me.

mgarstenauer:
I increased the minimum size of the `LineNumberMargin`. This ensures that there is enough space between the line numbers and any other text editor margins.

Together with the breakpoint margin it look like this:
image

mgarstenauer:
I have added a property `LineNumberMargin.MinWidthInDigits`. The width of the `LineNumberMargin` is now styleable from XAML. Example:

```xml
<AvalonEdit:TextEditor.Styles>
  <Style Selector="editing|LineNumberMargin">
    <Setter Property="MinWidthInDigits" Value="4" />
  </Style>
</AvalonEdit:TextEditor.Styles>
```

The defaults are as before to make sure that the changes in this PR don't mess up the layout of existing apps.

maxkatz6 (approved):

PR #421, opened 2024-05-02
Handling Selection get/set in TextInputMethodClient

### LINKED ISSUE #417: Repetitive input on Android
Each time you enter a new character, the previous text is inserted before it. Like if you want to write "hey", it will turn into "hhehey". This behavior starts again when the software keyboard is reset.
AndroidEdit

I found that the "Text" property of args in the "OnTextInput" for each new character has all the text entered (with the previous characters). 

.net8.0
Android API 31
Avalonia 11.0.10
AvaloniaEdit 11.0.6

Necromanika:
## What does the pull request do?
Handling selection change on `TextArea` and properly sync it with `TextInputMethodClient` callers.

## What is the current behavior?
Currently `TextInputMethodClient` mostly ignore selection `Set` and doesnt respect `TextEditor` indexes of columns and rows.

## What is the updated/expected behavior with this PR?
Working with IME

## How was the solution implemented (if it's not obvious)?
The solution was achieved by correctly using column and row indexes.
Also for word completion and correction work, the `_emptySelection` field was added, which is necessary to simulate changing the empty selection step by step (start and end) as it works in Avalonia `TextBox`.

## Fixed issues
Fixes #417

### DIFF
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml
@@ -3,7 +3,7 @@
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         MinWidth="0"
         MinHeight="300"
-        Width="950"
+        Width="1000"
         Title="AvaloniaEdit Demo"
         x:Class="AvaloniaEdit.Demo.MainWindow">
     <DockPanel>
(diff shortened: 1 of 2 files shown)


Necromanika:
@dotnet-policy-service agree

PR #433, opened 2024-06-20
Change Theme to a DropDown with flyout 

Natestah:
- Little bit of MVVP in here to enable bindings in the AXAML
- Drop down, allows me to look at specific themes, As I intend to leverage a little more of the info from the TextMateSharp JSON's

HELLO, I'm the author of BlitzSearch ( https://natestah.com ).  I'm making heavy use of AvaloniaEdit and there are things that I want to do with it.  This would be a small Ice Breaker change, would be awesome if the efforts could align.

### DIFF
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml
@@ -43,7 +43,7 @@
             <Button Name="insertSnippetBtn" Content="Insert Snippet" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
         </StackPanel>
         <StackPanel Name="StatusBar" Background="Purple" Height="25" DockPanel.Dock="Bottom" Orientation="Horizontal">
-            <TextBlock Foreground="White" Name="StatusText" Text="Ready" Margin="5 0 0 0" VerticalAlignment="Center" FontSize="12"/>
+            <TextBlock Name="StatusText" Text="Ready" Margin="5 0 0 0" VerticalAlignment="Center" FontSize="12"/>
         </StackPanel>
         <AvalonEdit:TextEditor Name="Editor"
                                FontFamily="Cascadia Code,Consolas,Menlo,Monospace"
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml
@@ -1,10 +1,12 @@
 ﻿<Window xmlns="https://github.com/avaloniaui"
         xmlns:AvalonEdit="using:AvaloniaEdit"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:vm="clr-namespace:AvaloniaEdit.Demo.ViewModels"
         MinWidth="0"
         MinHeight="300"
         Width="1000"
         Title="AvaloniaEdit Demo"
+        x:DataType="vm:MainWindowViewModel"
         x:Class="AvaloniaEdit.Demo.MainWindow">
     <DockPanel>
         <StackPanel Orientation="Horizontal"
@@ -23,7 +25,21 @@
             <Button Name="addControlBtn" Content="Add Button" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
             <Button Name="clearControlBtn" Content="Clear Buttons" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
             <ComboBox Name="syntaxModeCombo" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
-            <Button Name="changeThemeBtn" Content="Change theme" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
+            
+            <DropDownButton Content="Change theme" VerticalAlignment="Stretch" VerticalContentAlignment="Center">
+                <Button.Flyout>
+                    <Flyout>
+                        <ListBox ItemsSource="{Binding AllThemes }" SelectedItem="{Binding SelectedTheme}">
+                            <ListBox.DataTemplates>
+                                <DataTemplate x:DataType="vm:ThemeViewModel">
+                                    <TextBlock Text="{Binding DisplayName}"/>
+                                </DataTemplate>
+                            </ListBox.DataTemplates>
+                        </ListBox>    
+                    </Flyout>
+                </Button.Flyout>
+            </DropDownButton>
+            
             <Button Name="insertSnippetBtn" Content="Insert Snippet" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
         </StackPanel>
         <StackPanel Name="StatusBar" Background="Purple" Height="25" DockPanel.Dock="Bottom" Orientation="Horizontal">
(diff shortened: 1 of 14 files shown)


src/AvaloniaEdit.Demo/MainWindow.xaml
```diff
@@ -23,7 +25,21 @@
             <Button Name="addControlBtn" Content="Add Button" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
             <Button Name="clearControlBtn" Content="Clear Buttons" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
             <ComboBox Name="syntaxModeCombo" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
-            <Button Name="changeThemeBtn" Content="Change theme" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
+            
+            <DropDownButton Content="Change theme" VerticalAlignment="Stretch" VerticalContentAlignment="Center">
+                <Button.Flyout>
+                    <Flyout>
+                        <ListBox ItemsSource="{Binding AllThemes }" SelectedItem="{Binding SelectedTheme}">
```
maxkatz6, line 32: It can be MenuFlyout with ToggleType=Radio, in 11.1 Avalonia.

maxkatz6 (approved):

src/AvaloniaEdit.Demo/MainWindow.xaml
Natestah: The toggle type radio is pretty awesome, I'm using it in Blitz Search but sure if that's something we want with this library? since 11.1 is all marked rc1,beta, etc.

src/AvaloniaEdit.TextMate/AvaloniaEdit.TextMate.csproj
```diff
@@ -12,8 +12,9 @@
 
   <ItemGroup>
     <PackageReference Include="Avalonia" Version="$(AvaloniaVersion)" />
-    <PackageReference Include="TextMateSharp" Version="$(TextMateSharpVersion)" />
-    <PackageReference Include="TextMateSharp.Grammars" Version="$(TextMateSharpVersion)" />
+    <PackageReference Include="Avalonia.Themes.Fluent" Version="11.0.0" />
```
maxkatz6, line 15: Can we revert these changes from non-demo project?
Controls shouldn't depend on a single specific theme.

src/AvaloniaEdit.TextMate/TextMate.cs
```diff
@@ -61,17 +70,93 @@ public void SetGrammar(string scopeName)
...
 
-                GetOrCreateTransformer().SetTheme(_textMateRegistry.GetTheme());
+                var registryTheme = _textMateRegistry.GetTheme();
+                GetOrCreateTransformer().SetTheme(registryTheme);
 
                 _tmModel?.InvalidateLine(0);
 
                 _editorModel?.InvalidateViewPortLines();
+
+                if (Application.Current?.ApplicationLifetime is IClassicDesktopStyleApplicationLifetime
```
maxkatz6, line 96: It's a bad practice to depend on IClassicDesktopStyleApplicationLifetime.

src/AvaloniaEdit.TextMate/TextMate.cs
```diff
@@ -61,17 +70,93 @@ public void SetGrammar(string scopeName)
...
+                    {
+                        MainWindow: Window window
+                    })
+                {
+                    _guiColorDictionary = registryTheme.GetGuiColorDictionary();
+                    var themeName = theme.GetName();
+
+                    // Applies to the application on a whole.. Some might want to opt out of this for their own setups.
+                    if (applyWindowProperties)
```
maxkatz6, line 105: I am not sure if I understand these changes, but why would control change global theme?
If text editor has custom theming solution, it should be contained by this control.

src/AvaloniaEdit.TextMate/TextMate.cs
```diff
@@ -61,17 +70,93 @@ public void SetGrammar(string scopeName)
...
+                        _editor.LineNumbersForeground = _editor.Foreground;
+                    }
+                    
+                    AppliedTheme?.Invoke(this,this);
+                }
             }
+            
+            //These come out of CurrentHighlightRenderer sealed class. 
+            public static readonly Color CurrentHighlightRendererDefaultBackground = Color.FromArgb(22, 20, 220, 224);
+            public static readonly Color CurrentHighlightRendererDefaultBorder = Color.FromArgb(52, 0, 255, 110);
```
maxkatz6, line 159: These hardcoded colors should be avoided and removed in CurrentLineHighlightRenderer class as well. Colors should be defined by XAML themes. But that's for another day, I guess.

maxkatz6 (changes requested):

danipen:
@Natestah I tried the changes and found some issues:
* Once you open the demo window for the first time, the selection is not visible or it doesn't work.
* After changing the theme, the selection becomes visible.
* After changing the theme again, the selection color doesn't properly changes.

See the attached video for clarification:

https://github.com/AvaloniaUI/AvaloniaEdit/assets/501613/27b92e3c-6d86-491e-9318-fb2675dbeb3b

src/AvaloniaEdit.TextMate/TextMate.cs
Natestah: I at least got rid of the Duplicate definition,  but I agree, this should be some property up higher on the textview, not in the background highlight renderer. Just not today

src/AvaloniaEdit.TextMate/TextMate.cs
Natestah: Still getting my head around some of the Avalonia stuff, but this wasn't the place for this. it's now living in the Demo apps app.axaml.cs

src/AvaloniaEdit.TextMate/TextMate.cs
Natestah: removed..

Natestah:
I think Everything is up-to-date, thank you for time and patience. I'm still learning how to use GitHub and Avalonia

maxkatz6 (approved):
LGTM, thank you.

maxkatz6:
cc @danipen if you want to have another look

danipen:
I'm rethinking this a bit. 

We should let users use this functionality, but it shouldn't be mandatory. Some customers likely have custom background or selection colors, so we'd be breaking their styles. Setting styles for the text editor should be the user's responsibility. 

The TextMate API should be responsible of notifying about theme changes and providing the theme colors. We might also include constants to help the users to easily get the colors. Let me rethink this a bit and I'll do a small refactor.

src/AvaloniaEdit.TextMate/TextTransformation.cs
```diff
@@ -18,7 +18,7 @@ public class ForegroundTextTransformation : TextTransformation
         public Action<Exception> ExceptionHandler { get; set; }
         public int ForegroundColor { get; set; }
         public int BackgroundColor { get; set; }
-        public int FontStyle { get; set; }
+        public FontStyle FontStyle { get; set; }
```
maxkatz6, line 21: I haven't noticed before, but changing public property type is an ABI breaking change. Might not be critical, but needs to be carefully considered.

danipen: Yes, this is an API breaking from TextMateSharp ...

PR #440, opened 2024-07-22
fix: System.NullReferenceException on Undo

### LINKED ISSUE #425: System.NullReferenceException on Undo
## How to simulate:

- Clone latest master 88625c4491af8ca6fa9127872ae1f1099fd06392
- Open `AvaloniaEdit.sln`
- Run `AvaloniaEdit.Demo`
- Go to line 62
- Press `Enter` key
- Click on Button `Insert Snipet` 
- Edit name in `aaa` and press `Enter` key for confirm the change
- Press `CTRL+Z` four times
- The exception will be thrown

workgroupengineering:
fixes #425

### DIFF
--- a/src/AvaloniaEdit.TextMate/TextEditorModel.cs
+++ b/src/AvaloniaEdit.TextMate/TextEditorModel.cs
@@ -7,6 +7,12 @@ using AvaloniaEdit.Rendering;
 
 using TextMateSharp.Model;
 
+#if NET6_0_OR_GREATER
+using Math = System.Math;
+#else
+using Math = AvaloniaEdit.TextMate.Compatibility.Math;
+#endif
+
 namespace AvaloniaEdit.TextMate
 {
     public class TextEditorModel : AbstractLineList, IDisposable
--- a/src/AvaloniaEdit.TextMate/TextEditorModel.cs
+++ b/src/AvaloniaEdit.TextMate/TextEditorModel.cs
@@ -168,7 +168,7 @@ namespace AvaloniaEdit.TextMate
 
             try
             {
-                InvalidateLineRange(_invalidRange.StartLine, _invalidRange.EndLine);
+                InvalidateLineRange(_invalidRange.StartLine < _documentSnapshot.LineCount ? _invalidRange.StartLine : _document.LineCount - 1, _invalidRange.EndLine < _documentSnapshot.LineCount ? _invalidRange.EndLine : _documentSnapshot.LineCount - 1);
             }
             finally
             {
--- a/src/AvaloniaEdit.TextMate/TextEditorModel.cs
+++ b/src/AvaloniaEdit.TextMate/TextEditorModel.cs
@@ -168,7 +168,9 @@ namespace AvaloniaEdit.TextMate
 
             try
             {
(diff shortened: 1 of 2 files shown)


workgroupengineering:
Do I need to make any other changes? I hope the merge and release takes place soon.

src/AvaloniaEdit.TextMate/TextEditorModel.cs
```diff
@@ -168,7 +168,7 @@ void DocumentOnUpdateFinished(object sender, EventArgs e)
 
             try
             {
-                InvalidateLineRange(_invalidRange.StartLine, _invalidRange.EndLine);
+                InvalidateLineRange(_invalidRange.StartLine < _documentSnapshot.LineCount ? _invalidRange.StartLine : _document.LineCount - 1, _invalidRange.EndLine < _documentSnapshot.LineCount ? _invalidRange.EndLine : _documentSnapshot.LineCount - 1);
```
danipen, line 171: Could you add a method to clamp the line number for better readability?
```suggestion
    int startLine = Math.Clamp(_invalidRange.StartLine, 0, _documentSnapshot.LineCount - 1);
    int endLine = Math.Clamp(_invalidRange.EndLine, 0, _documentSnapshot.LineCount - 1);
    InvalidateLineRange(startLine, endLine);
```

danipen (changes requested):

src/AvaloniaEdit.TextMate/TextEditorModel.cs
danipen: Anyway, the changes are ok. 

I should probably protect the NRE later in TextMateSharp library, just in case the consumer passes invalid line numbers as in this case. Thanks for the PR.

danipen (approved):
LGTM

danipen:
It seems that the Azure Pipelines [task](https://dev.azure.com/AvaloniaUI/AvaloniaUI/_build/results?buildId=50552&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&s=96ac2280-8cb4-5df5-99de-dd2da759617d&t=ea74d3ec-af51-57bf-65e6-f50dd234a1e5&l=123) `XXX` is failing in a consistent way. Any ideas?

```
##[warning].NET 5 has some compatibility issues with older Nuget versions(<=5.7), so if you are using an older Nuget version(and not dotnet cli) to restore, then the dotnet cli commands (e.g. dotnet build) which rely on such restored packages might fail. To mitigate such error, you can either: (1) - Use dotnet cli to restore, (2) - Use Nuget version 5.8 to restore, (3) - Use global.json using an older sdk version(<=3) to build
##[error]Error: The process 'C:\Program Files\dotnet\dotnet.exe' failed with exit code 1
##[error]An error occurred while trying to pack the files.
Info: Azure Pipelines hosted agents have been updated and now contain .Net 5.x SDK/Runtime along with the older .Net Core version which are currently lts. Unless you have locked down a SDK version for your project(s), 5.x SDK might be picked up which might have breaking behavior as compared to previous versions. You can learn more about the breaking changes here: https://docs.microsoft.com/en-us/dotnet/core/tools/ and https://docs.microsoft.com/en-us/dotnet/core/compatibility/ . To learn about more such changes and troubleshoot, refer here: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#troubleshooting
Finishing: Pack AvaloniaEdit
```

Any ideas?

workgroupengineering:
I see. Net standard 2.0 does not support Clamp. I fix it tomorrow.

PR #443, opened 2024-07-28
Bump TextMateSharp version

Blquinn:
Bumps TextMateSharp version to latest, which allows usage of changes from https://github.com/danipen/TextMateSharp/pull/67

### DIFF
--- a/src/AvaloniaEdit.TextMate/AvaloniaEdit.TextMate.csproj
+++ b/src/AvaloniaEdit.TextMate/AvaloniaEdit.TextMate.csproj
@@ -12,8 +12,8 @@
 
   <ItemGroup>
     <PackageReference Include="Avalonia" Version="$(AvaloniaVersion)" />
-    <PackageReference Include="TextMateSharp" Version="1.0.58" />
-    <PackageReference Include="TextMateSharp.Grammars" Version="1.0.58" />
+    <PackageReference Include="TextMateSharp" Version="1.0.59" />
+    <PackageReference Include="TextMateSharp.Grammars" Version="1.0.59" />
   </ItemGroup>
 
 </Project>


Blquinn:
I noticed while working on this change that if I included the package reference for the updated TextMatesSharp:

```xml
<PackageReference Include="TextMateSharp.Grammars" Version="1.0.59" />
```

in my project's csproj that syntax highlighting silently broke.

Maybe someone has thoughts on why/how that might happen?

danipen:
> in my project's csproj that syntax highlighting silently broke.

Do you mean the IDE's syntax highlighting?

Blquinn:
> Do you mean the IDE's syntax highlighting?

No, the TextEditor widget's syntax highlighting stopped working.

Blquinn:
I should add that I tested this before PR'ing the Grammars change by branching master of this repo and bumping the Grammars package. Syntax highlighting was working.

Numpsy:
> I noticed while working on this change that if I included the package reference for the updated TextMatesSharp:
> 
> ```
> <PackageReference Include="TextMateSharp.Grammars" Version="1.0.59" />
> ```
> 
> in my project's csproj that syntax highlighting silently broke.
> 
> Maybe someone has thoughts on why/how that might happen?

I got the same when I tried that to get the System.Text.Json update - Visual Studio says this:

image

I haven't had time to look at anything beyond that.

Numpsy:
I'd guess caused by https://github.com/danipen/TextMateSharp/pull/59 - 

image

Blquinn:
Ah, I may be able to take a look at that tonight. It's a little concerning that the stack trace gets buried... maybe that's the configuration of my app though.

Blquinn:
I tried checking out AvaloniaEdit `11.0.6` and bumping the TextMateSharp version to `1.0.59`, which won't compile.

As @Numpsy showed there's a breaking change in TextMateSharp where some int fields are changed into the FontStyle enum. So the current version of AvaloniaEdit on Nuget isn't compatible with versions of  TextMateSharp >  `1.0.56`.

I'm pretty confused why I don't see an exception being thrown. Perhaps dotnet is happy to do this conversion between ints and the enum when I'm using pre-compiled byte-code or something.

Things do compile and the demo runs normally against master (This PR) though. I think there just needs to be a newer version of AvaloniaEdit released to resolve the issue.

danipen:
@Takoooooo please could you do that?

maxkatz6:
@danipen I can do a new release. But should I include this PR? If you want it to be compatible with a new text mate.

danipen:
Well ... the PR that fixes the issue is this one https://github.com/AvaloniaUI/AvaloniaEdit/pull/441.

Also note that the TextMateSahrp API changed some versions ago (at version 1.0.56). The API change is minimal (it's just changing a int type by an enum) so I think we're ok going with it and merging this PR.

PR #444, opened 2024-07-29
Fix customMargin renders incorrectly when scrolling

### LINKED ISSUE #442: CustomMargin renders incorrectly when scrolling
https://github.com/AvaloniaUI/AvaloniaEdit/blob/934f022406334b95813ecb9040b9e8864b81e1ac/src/AvaloniaEdit.Demo/CustomMargin.cs#L115

is fixed with:
`double y = visualLine.VisualTop - TextView.VerticalOffset + visualLine.Height / 2;`

clicked-line10-no-scroll
clicked-line10-scrolled-down
clicked-line10-scrolled-down-fixed

danipen:
Fixes #442.

- Fixed custom margin rendering visuals in the wrong position when scrolling.
- Added `InvalidateVisual()` when the pointer exits to hide the hovered marker.
- Cleared markers when changing the document.

### DIFF
--- a/src/AvaloniaEdit.Demo/CustomMargin.cs
+++ b/src/AvaloniaEdit.Demo/CustomMargin.cs
@@ -4,6 +4,7 @@ using Avalonia;
 using Avalonia.Input;
 using Avalonia.Media;
 using Avalonia.Media.Immutable;
+using AvaloniaEdit.Document;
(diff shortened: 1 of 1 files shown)


mgarstenauer:
I can reproduce #442. Not sure how I missed that.

Thanks for making the fixes. Changes look good.

PR #446, opened 2024-08-05
Mouse ContextFlyout commands fix in AvaloniaEdit.Demo

Mop1398:
hello 

this is a PR fixing Commands of Mouse ContextMenu for AvaloniaEdit.Demo

used ContextFlyout instead of ContextMenu
previous actions which were not working:

Screenshot 2024-08-05 123041

now these Items working (it's not ReactiveUI, basic MVVM):

Screenshot 2024-08-05 124512

couldn't find a better way to call the Commands from Demo. but this seems working just fine .in that case let me know here,

PS:   it's my very first PR 😅, hope it helps and more importantly hope i've done it the right way

### DIFF
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml.cs
@@ -51,7 +51,8 @@ namespace AvaloniaEdit.Demo
             _textEditor.HorizontalScrollBarVisibility = Avalonia.Controls.Primitives.ScrollBarVisibility.Visible;
             _textEditor.Background = Brushes.Transparent;
             _textEditor.ShowLineNumbers = true;
-            _textEditor.ContextMenu = new ContextMenu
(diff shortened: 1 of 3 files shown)


src/AvaloniaEdit.Demo/MainWindow.xaml.cs
```diff
@@ -51,15 +51,16 @@ public MainWindow()
             _textEditor.HorizontalScrollBarVisibility = Avalonia.Controls.Primitives.ScrollBarVisibility.Visible;
             _textEditor.Background = Brushes.Transparent;
             _textEditor.ShowLineNumbers = true;
-            _textEditor.ContextMenu = new ContextMenu
+            // The ContextMenu has been defined in MainWindow.xaml
```
danipen, line 54: You can remove the commented code for clarity.

danipen (changes requested):

danipen (approved):

PR #450, opened 2024-08-28
Do not run the element generators for performance reasons

danipen:
Having documents with very long lines, 1M columns or so (this can happen for minified json or javascript) caused the UI to stop running.

### DIFF
--- a/src/AvaloniaEdit/Rendering/VisualLine.cs
+++ b/src/AvaloniaEdit/Rendering/VisualLine.cs
@@ -166,6 +166,13 @@ namespace AvaloniaEdit.Rendering
             LastDocumentLine = FirstDocumentLine;
             var askInterestOffset = 0; // 0 or 1
 
+            // for long lines, do not run the element generators for performance reasons
+            if (lineLength > LENGTH_LIMIT)
+            {
+                _elements.Add(new VisualLineText(this, lineLength));
+                return;
+            }
+
             while (offset + askInterestOffset <= currentLineEnd)
             {
                 var textPieceEndOffset = currentLineEnd;


Gillibald:
Is there a way to process some parts of the document later? For example when the user scrolls into a text region that wasn't processed before? The question is shouldn't we trim the exceeding part and therefore never reach the limit when generators are running?

danipen:
Well, the ideal solution, as you've pointed out several times, would be to implement horizontal virtualization for lines. However, that would involve significantly more changes. This PR is more of a workaround specifically for minified files, like those in a single line (often JSON or JavaScript). By disabling the generators (which match email addresses and other features) for long lines, we can avoid the UI hanging. While editing isn't totally smooth, it’s much better than the previous situation where running the generators took forever. Now, generating the line only takes a few milliseconds, which is acceptable.

Gillibald (approved):

KrzysztofDusko:
i worth to mention that this change disable all additional line generators.. For example as mitigation for slow performance for long lines i have this VisualLineElement Generator.. 
Probably soem, easy method to force some generators to work would be nice

```
public sealed class TruncateLongLines : VisualLineElementGenerator
{
    const int maxLength = 2000;
    const string ellipsis = "...";
    const int charactersAfterEllipsis = 100;

    public override int GetFirstInterestedOffset(int startOffset)
    {
        DocumentLine line = CurrentContext.VisualLine.LastDocumentLine;
        if (line.Length > maxLength)
        {
            int ellipsisOffset = line.Offset + maxLength - charactersAfterEllipsis - ellipsis.Length;
            if (startOffset <= ellipsisOffset)
                return ellipsisOffset;
        }
        return -1;
    }

    public override VisualLineElement ConstructElement(int offset)
    {
        return new FormattedTextElement(ellipsis, CurrentContext.VisualLine.LastDocumentLine.EndOffset - offset - charactersAfterEllipsis);
    }
}
```

danipen:
VSCode always truncate long lines so an option could be including your TruncateLongLines generator  as a built in generator and always run it before running other generators... What do you guys think?

KrzysztofDusko:
> VSCode always truncate long lines so an option could be including your TruncateLongLines generator as a built in generator and always run it before running other generators... What do you guys think?

probably folding generators should still have highest priority?
https://github.com/AvaloniaUI/AvaloniaEdit/blob/master/src/AvaloniaEdit/Folding/FoldingManager.cs#L337

PR #414, opened 2024-04-17
Fix: #413 CompletionWindow doesn't work with mouse click

mgarstenauer:
Fixes #413.

Issue #413, opened 2024-04-16
CompletionWindow doesn't work with mouse click


NickRimmer:
Hey community,

I found strange behavior. When `CompletionWindow` opened, you can select variants by keyboard up/down and press enter - everything working as expected. But if you will open it and select option by mouse - it will just close `CompletionWindow` and do nothing.

How can we make it works by mouse selection as well?

## How to reproduce
- clone AvaloniaEdit master branch
- run `AvalolinaEdit.Demo` project
- in some place try to write something like `asd.` to show `CompletionWindow`
- click to any of items by mouse

![image](https://github.com/AvaloniaUI/AvaloniaEdit/assets/1754265/7d5ecc18-24dd-4c46-bc9e-260c24f1aecb)

mgarstenauer:
I looked at the code and found this TODO:

https://github.com/AvaloniaUI/AvaloniaEdit/blob/adbe72b8023207270d30c16beb4fadcd6d598023/src/AvaloniaEdit/CodeCompletion/CompletionList.cs#L190

Shouldn't be too hard to fix. Preparing a PR now...

What is the expected behavior? Do we want to insert the item on a single-click or a double-click? Most modern editors insert items with a single-click. IIRC, the original AvalonEdit used a double-click. I am going to implement single-click for now.

NickRimmer:
Thank you so much for quick review @danipen. 
@MarchingCube thank you for great fix 🎉

gebodal:
Apologies for being difficult, resurrecting an old issue, but it would be great if the single/double click was configurable somehow. There doesn't seem to be a particularly clean way of overloading these classes to change the behaviour, as the `AddHandler(...)` is embedded in the class constructor, so an option would be great.

mgarstenauer:
What would be the motivation for using double-click instead of single-click?

gebodal:
The motivation for me would be that the ListBox is a SelectingItemsControl, and the fact that the selection works with the arrow keys but not the mouse pointer breaks the affordance (had to look up the technical term there) of the control. It also doesn't seem consistent across the various interactions. And, as the action now triggers on PointerPressed rather than PointerReleased or DoubleTapped, the action occurs sooner than expected, which slightly diminishes recoverability for user input.

As an example, Visual Studio uses a single-click for selecting items in the autocomplete dropdown, and double-click for activating a given autocomplete option.

mgarstenauer:
[...]

PR #451, opened 2024-08-29
Add option to insert completion item on PointerPressed/PointerReleased/DoubleTapped

mgarstenauer:
This PR adds an options to choose whether completion items are inserted on `PointerPressed`, `PointerReleased` or `DoubleTapped`.

This is an improvement on #414 and addresses the feedback by @gebodal mentioned on issue #413.

In addition, this fixes an issue where first click in the `CompletionWindow` is swallowed by the `PopupWithCustomPosition` tooltip.

### DIFF
--- a/src/AvaloniaEdit/CodeCompletion/CompletionAcceptAction.cs
+++ b/src/AvaloniaEdit/CodeCompletion/CompletionAcceptAction.cs
@@ -1,0 +1,25 @@
+﻿namespace AvaloniaEdit.CodeCompletion;
+
+/// <summary>
+/// Defines the pointer action used to request the insertion of a completion item.
+/// </summary>
+public enum CompletionAcceptAction
+{
+    /// <summary>
+    /// Insert the completion item when the pointer is pressed. (This option makes the completion
+    /// list behave similar to the completion list in Visual Studio Code.)
+    /// </summary>
+    PointerPressed,
+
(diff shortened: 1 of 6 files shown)


src/AvaloniaEdit/CodeCompletion/CompletionAcceptAction.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace AvaloniaEdit.CodeCompletion;
+
+/// <summary>
+/// Defines the pointer action used to request the insertion of a completion item.
+/// </summary>
+public enum CompletionAcceptAction
```
maxkatz6, line 6: Feels like it could be a flags enum, with a possibility to set `None` option at least.

mgarstenauer: The current options `PointerPressed`/`PointerReleased`/`DoubleTapped` are mutually exclusive and can't be combined as flags.

I am not sure if we want to allow `None`. That would prevent users from using the mouse for insertions. I suggest to only add `None` if there is practical use case.

The current defaults (`TextEditorOptions.CompletionAcceptAction == CompletionAcceptAction.PointerPressed`) ensure that the behavior in existing apps is unchanged.

maxkatz6 (approved):

PR #464, opened 2024-11-07
Add watermark support

danipen:
Support watermark text when the text in the text area is empty:

https://github.com/user-attachments/assets/4b28fc6b-d577-40e0-9e84-4869c92b6d1a


danipen:
@Gillibald I have a question about IME here. 

In the Avalonia's [TextBox implementation](https://github.com/AvaloniaUI/Avalonia/blob/9cecb90ba1e681d1783e3a89db4b8d860859550f/src/Avalonia.Themes.Fluent/Controls/TextBox.xaml#L155) you’re checking the TextPresenter’s preedit text to determine if we should display the watermark. Should we do the same for AvaloniaEdit? I see that the InputMethodClient in the TextArea [doesn’t support preedit](https://github.com/AvaloniaUI/AvaloniaEdit/blob/f2e6452c25da12a28bce1074f468e20789f3b837/src/AvaloniaEdit/Editing/TextArea.cs#L1170), so probably is not needed …

Gillibald (approved):

PR #470, opened 2024-11-20
Avoid full redraw when selection brush changes

mgarstenauer:
This PR makes a small fix to prevent unnecessary redraws of the `TextView`.

## Description of change

The `TextArea` calls `TextView.Redraw()` when the selection style properties (`SelectionBrush`, `SelectionBorder`, `SelectionCornerRadius`) change. But in this case it is not necessary to do a full redraw. It is sufficient to redraw only the `Selection` layer.

## Problem with original code

A full redraw is a harmful because it interferes with the `SelectionMouseHandler` in the following scenario:

I am using a custom theme to override the selection brush depending on whether the `TextEditor` has focus.

```xml
<ControlTheme x:Key="{x:Type editing:TextArea}" TargetType="editing:TextArea">
  <Setter Property="SelectionBrush" Value="Gray" />
  ...
  <Style Selector="^:focus-within">
    <Setter Property="SelectionBrush" Value="Blue" />
  </Style>
  </ControlTheme>
</ResourceDictionary>
```

This creates the following problem: When `TextEditor` is clicked and receives focus, the text from the last position up to the new caret position is selected.

Expected behavior: Clicking the `TextEditor` should not create a selection. Only click-and-drag should create a new text selection.


mgarstenauer:
FYI: The CI builds currently fail with

```
error NU1903: Warning As Error: Package 'System.Text.Json' 8.0.4 has a known high severity vulnerability
```

which is not caused by any of my change. I have removed `<TreatWarningsAsErrors>false</TreatWarningsAsErrors>` locally to get around the issue.

maxkatz6 (approved):

PR #487, opened 2025-02-15
Feature: Allow loading grammar files from path

LingFeng-bbben:


src/AvaloniaEdit.TextMate/TextMate.cs
```diff
@@ -63,6 +63,15 @@ public void SetGrammar(string scopeName)
                 _editor.TextArea.TextView.Redraw();
             }
 
+            public void SetGrammarFile(string path)
```
danipen, line 66: Can you please extract a common method that receives a `IGrammar` and it's used by both `SetGrammar` and `SetGrammarFile` to avoid duplicated logic?

danipen (changes requested):
Thanks for your contribution! Please see some comments....

LingFeng-bbben:
Sorry for pushing my own purposed change to this wrong branch.... Will make a new pr for this

PR #489, opened 2025-02-16
Adjust caret rectangle for scroll offset

### LINKED ISSUE #352: IME BUG：After scrolling, the Chinese input method does not work properly
I have a large line of text and cannot type Chinese properly after scrolling

realybin:
Fixes #352
## Before this pr
before
## After this pr
after


Gillibald (approved):
LGTM

realybin:
I meet some issues while my mate test this pr. I'm investigating it.

realybin:
image
The same problem occurs with the TextBox in the search box, but this existed before I submitted this pr, and after modifying it, this pr didn't make this function correctly, and it's common sense that the TextBox should be displayed correctly as expected. Why is this? What are your thoughts on this please? @Gillibald @danipen

d3ara1n:
Any updates? I have ran into the same problem :sad:

PR #496, opened 2025-03-28
Fix: Rectangular selection ignores scroll offset

### LINKED ISSUE #472: Rectangular selection shows interval at line end when text is longer than screen
1. step1: input following code, the code length needs to be longer than window width
`select arandomtable1.id,
       arandomtable1.relative_flag,
       arandomtable1.randomcolumn1
from arandometable1
         join arandomtable2 on randomtable2.relative_id = arandomtable1.id
    and arandometable.delete_flag = false
         join arandomtable3 on randomtable2.relative_id = arandomtable3.id and
                               now() between arandomtable3.effective_start and arandomtable3.effective_end
         join arandomtable4 on randomtable2.relative_id = arandomtable4.id and arandometable.delete_flag = false
    and arandomtable2.another_relative_id = arandomtable4.id and
                               arandomtable4.radom_column in ('random1', 'random2', 'random3', 'random5')`
2. step2:start rectangular selection and continue draging to the right side when the horizontal scrollbar shows
3. step3:the interval appears in the line end
image

mgarstenauer:
Fixes #472.

When a selection extends beyond the visual line into the "virtual space" then the `BackgroundGeometryBuilder` creates two segments per line. One for the visible text, and one for the virtual space.

The segment for the virtual space didn't apply the scroll offset. This is now fixed.

### DIFF
--- a/src/AvaloniaEdit/Rendering/BackgroundGeometryBuilder.cs
+++ b/src/AvaloniaEdit/Rendering/BackgroundGeometryBuilder.cs
@@ -260,10 +260,13 @@ namespace AvaloniaEdit.Rendering
 					} else {
 						right = visualLine.GetTextLineVisualXPosition(lastTextLine, segmentEndVc);
 					}
+
+					left -= scrollOffset.X;
+					right -= scrollOffset.X;
 					Rect extendSelection = new Rect(Math.Min(left, right), y, Math.Abs(right - left), line.Height);
 					if (lastRect != default) {
 						if (extendSelection.Intersects(lastRect)) {
-							lastRect.Union(extendSelection);
+							lastRect = lastRect.Union(extendSelection);
 							yield return lastRect;
 						} else {
 							// If the end of the line is in an RTL segment, keep lastRect and extendSelection separate.


src/AvaloniaEdit/Rendering/BackgroundGeometryBuilder.cs
```diff
@@ -260,10 +260,13 @@ private static IEnumerable<Rect> ProcessTextLines(TextView textView, VisualLine
...
 						right = visualLine.GetTextLineVisualXPosition(lastTextLine, segmentEndVc);
 					}
+
+					left -= scrollOffset.X;
+					right -= scrollOffset.X;
 					Rect extendSelection = new Rect(Math.Min(left, right), y, Math.Abs(right - left), line.Height);
 					if (lastRect != default) {
 						if (extendSelection.Intersects(lastRect)) {
-							lastRect.Union(extendSelection);
+							lastRect = lastRect.Union(extendSelection);
```
mgarstenauer, line 269: This is a second, unrelated fix where the selection is not properly extended because `Union` is use wrong.

danipen:
Thanks for the contribution! I noticed that the upstream sources [don't have this fix](https://github.com/icsharpcode/AvalonEdit/blob/30cad99ce905412ed5f5e847e3c00e72e69aee77/ICSharpCode.AvalonEdit/Rendering/BackgroundGeometryBuilder.cs#L264). Do you know if they’re also experiencing this issue? It might be worth checking just in case.

mgarstenauer:
`AvalonEdit` has the same problem. I'll provide a separate PR for upstream.

PR #498, opened 2025-03-29
Fix: Buttons not drawn when unsorted

mgarstenauer:
When clicking `Add Button` in the `AvaloniaEdit.Demo`, some buttons are never drawn.

### Cause
The `ElementGenerator` uses `BinarySearch` to find the first offset with a `Button`. But the list is never actually sorted in the demo.

### Solution
Sort the list after each insert.


mgarstenauer:
A cleaner solution would be to refactor it: Make the list private and add `ElementGenerator.Add/Remove/Clear` methods. Sort automatically in `ElementGenerator.Add`.

I opted for the one-line fix.

PR #500, opened 2025-03-29
Fix: Update the selection when scrolling

### LINKED ISSUE #426: Scrolling while the left mouse button is held to select text doesn't update selection
## Steps

1. Press and hold the left mouse button (without any modifiers)
2. Begin scrolling either with the mouse scroller or with some other method (trackpad, etc.)
3. Do not move the mouse pointer

## Expected

The selection span is updated accordingly.

## Actual

The selection span is not updated. It will only be updated after the pointer moves, triggering a `PointerMoved` event.

mgarstenauer:
This PR adds a handler for `ScrollOffsetChanged` to the `SelectionMouseHandler`.

Fixes #426.


mgarstenauer:
I rebased the branch on latest master and resolved the merge conflict.

PR #515, opened 2025-06-06
Optimize heap allocations in CharRope.ToString

YoshihiroIto:
Hi team,

I noticed that `CharRope.ToString` in AvaloniaEdit performs a large number of heap allocations, especially with large strings.

This pull request optimizes `CharRope.ToString` to use `string.Create` on .NET 6 and newer targets, which significantly improves performance. For older versions, the existing logic is kept as a fallback to ensure backward compatibility.

To implement this, I've added `Span<T>` overrides to several internal methods or updated them to use `Span<T>`.

This optimization should lead to reduced memory usage and contribute to better overall application performance.

I'd appreciate it if you could take a look. Thank you!


maxkatz6:
Thanks!

PR #537, opened 2025-09-17
Remove hardcoded line height factor

### LINKED ISSUE #526: hard coded constant in calculation of lineHeight casues incorrect rendering when fractional line heights are not appropriate.
My target environment needs to have zero space between my lines, but AvaloniaEdit has a hard coded factor of 1.35 multiplied times the font height.

This causes miscalculations in the height of the line causing it render incorrectly and messes with selection, keyboard navigation etc.

Image

when I remove the *1.35 in a private branch it renders 100% accurately:

Image

**The issue is in VisualLineTextParagraphProperties**
```c#
sealed class VisualLineTextParagraphProperties : TextParagraphProperties
{
...
	public override double LineHeight => DefaultTextRunProperties.FontRenderingEmSize * 1.35;
..}
```
LineHeight is hard coded to be target font size * 1.35 scaling factor.  This is a reasonable default, but there is no way to change it without forking the code and having to push an alternative nuget package. 

**Proposed solution**
there need to be SOME way of overriding this constant without having to fork the entire code base. Suggestion is to add top level property on TextEditor which is an option to control the factor and have it default to 1.35.

something like this:
```xaml
        <AvalonEdit:TextEditor Name="Editor"  ShowLineNumbers="True" LineSpacingFactor="1"/>
```

mgarstenauer:
Remove the hardcoded line height factor in `VisualLineTextParagraphProperties`.
Use `NaN` to determine line height automatically from font characteristics.

Fixes #526

The resulting line spacing is a little bit tighter than before, but more in line with the font characteristics. 

Here is before (left, factor 1.35) and after (right, using NaN).

image

As mentioned in #526, an additional option `TextEditorOptions.LineSpacingFactor` would be nice to have. But this can be a separate change.

### DIFF
--- a/src/AvaloniaEdit/Rendering/VisualLineTextParagraphProperties.cs
+++ b/src/AvaloniaEdit/Rendering/VisualLineTextParagraphProperties.cs
@@ -34,7 +34,7 @@ namespace AvaloniaEdit.Rendering
 
 		public override FlowDirection FlowDirection => FlowDirection.LeftToRight;
 		public override TextAlignment TextAlignment => TextAlignment.Left;
-		public override double LineHeight => DefaultTextRunProperties.FontRenderingEmSize * 1.35;
+		public override double LineHeight => double.NaN;
 		public override bool FirstLineInParagraph => firstLineInParagraph;
 		public override TextRunProperties DefaultTextRunProperties => defaultTextRunProperties;
 


maxkatz6:
@danipen @Gillibald you previously changed this const to a current form, do you remember why it was there? Was it a hack for another issue or lack or LineSpacingFactor?

It seems to be intentional here https://github.com/AvaloniaUI/AvaloniaEdit/pull/216

mgarstenauer:
Rendering of custom controls seems broken in current master (bad alignment). 
With this PR it is working again.

Screenshot: ❌ current master (left), ✔️ this PR (right)

image


Inserting emojis is also not working as expected in current master. When I insert an emoji the text line shifts up by ~1 px. It works when using NaN.

Screenshot: ❌ current master (left), ✔️ this PR (right)

image

(All tests using Avalonia 11.3.0 on Windows 11.)

Gillibald:
I can't remember why this factor was introduced. The original file looks like so, and ideally, we should take that: https://github.com/icsharpcode/AvalonEdit/blob/master/ICSharpCode.AvalonEdit/Rendering/VisualLineTextParagraphProperties.cs

Gillibald (approved):

danipen:
@mgarstenauer. Thank you for your contribution! The PR works well. I noticed the lines look a bit compressed compared to other editors—Visual Studio, VS Code, or Sublime usually have slightly more spacing. The current compression slightly reduces readability, so if you have a chance, could you adjust the spacing to align with that style?

mgarstenauer:
@danipen I'll have a look.

JetBrains Rider also has `Line Height` setting and uses a more relaxed `1.1` line spacing by default. Would be definitely nice to have in AvaloniaEdit.

danipen:
Just to clarify, that value would act as a line height multiplier, so setting it to `1` would leave the line height unchanged. Your plan is to use `1.1` as the default value, correct?

mgarstenauer:
Yes, it should act as a line height multiplier.

Examples: 
- `1` ... same a `NaN`, determined by font characteristics.
- `1.1` ... increases line height by 10%.

I will adjust the default value so that it matches the previous look. (Similar to defaults in Visual Studio.)

PR #539, opened 2025-09-29
Add TextEditorOptions.LineHeightFactor

mgarstenauer:
This is a follow up to #537.

The PR adds `TextEditorOptions.LineHeightFactor`. 
The `LineHeightFactor` is only applied to the default text.
The default value is ~1.0455, which matches the Visual Studio text editor on my machine. 

Additional changes:
- The inline button in the demo is aligned with the text using `TextBlock.SetBaselineOffset`.
- New: Inline UI elements can now use `VerticalAlignment` within a line.
- Fix: `CalcCaretOverstrikeRectangle` was using the wrong `VisualYPosition`.

### Screenshots

Left: Visual Studio, right: AvaloniaEdit `LineHeightFactor = 1.0455` (default)
image

Here is AvaloniaEdit with inline button, emoji, special chars, word wrap, caret, line indicator. 
image

Same with multiline selection.
image

Here is the alignment of the code completion window.
image

Here with `LineHeightFactor = 2.0`.
image

### DIFF
--- a/src/AvaloniaEdit.Demo/MainWindow.xaml
+++ b/src/AvaloniaEdit.Demo/MainWindow.xaml
@@ -26,7 +26,6 @@
             <Button Name="addControlBtn" Content="Add Button" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
             <Button Name="clearControlBtn" Content="Clear Buttons" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
             <ComboBox Name="syntaxModeCombo" VerticalAlignment="Stretch" VerticalContentAlignment="Center"/>
-            
             <DropDownButton Content="Change theme" VerticalAlignment="Stretch" VerticalContentAlignment="Center">
                 <Button.Flyout>
                     <Flyout>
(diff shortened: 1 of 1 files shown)


src/AvaloniaEdit/TextEditorOptions.cs
```diff
@@ -719,5 +719,30 @@ public CompletionAcceptAction CompletionAcceptAction
                 }
             }
         }
+
+        // The default LineHeightFactor matches the line height in the Visual Studio text editor.
+        private const double DefaultLineHeightFactor = 1.0455;
```
danipen, line 724: I tested the changes and I still see the text quite compressed using the default `1.0455 `value. I suggest using a value that ensures the text is displayed in the same height before #537.
```suggestion
        private const double DefaultLineHeightFactor = 1.16;
```

See attached image 
image

danipen (changes requested):

danipen (approved):
LGTM

danipen:
@mgarstenauer I added a numeric up/down control to the demo app so it’s easier to test different `LineHeightFactor` values. Thanks again for your contribution!

tomlm:
This fixed a slew of bugs with Consolonia and AvaloniaEdit. When are you thinking of releasing?

PR #546, opened 2025-11-20
Adjust caret rectangle calculations to accomodate upcoming Avalonia versions

danipen:
This PR updates the caret rectangle calculation in `Caret.cs` to use `VisualYPosition.LineTop` and `VisualYPosition.LineBottom` instead of `TextTop` and `TextBottom`.

This is required for Avalonia `11.3.8` and later, where font metrics changed and `LineTop/LineBottom` no longer match `TextTop/TextBottom`.

With the current Avalonia version (`11.0.0`), this change has no effect because `LineTop` and `TextTop` still return the same values. The update simply prepares the code to behave correctly once we migrate to `11.3.8` or newer.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,7 +4,7 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AvaloniaVersion>11.0.0</AvaloniaVersion>
     <AvaloniaSampleVersion>11.3.0</AvaloniaSampleVersion>
-    <TextMateSharpVersion>1.0.66</TextMateSharpVersion>
+    <TextMateSharpVersion>2.0.2</TextMateSharpVersion>
     <VersionSuffix>beta</VersionSuffix>
 
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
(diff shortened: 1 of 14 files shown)


grokys:
Out of interest, why does AvaloniaEdit now need 11.3.9?

danipen:
> Out of interest, why does AvaloniaEdit now need 11.3.9?

Strictly speaking it doesn’t really need `11.3.9`. We updated our codebase to Avalonia `11.3.9` (we’re now on `11.3.11`), and during that upgrade we detected some text misalignment in AvaloniaEdit caused by changes in font metrics introduced in version (`11.3.9`).

To fix those issues, I pushed a PR with the necessary adjustments and, as part of that, bumped the Avalonia version in the AvaloniaEdit repo. So the changes are actually needed for any Avalonia version >= `11.3.9`, not just that exact one. 

The PR probably doesn’t explain this very clearly, sorry about that.

grokys:
It's just that this change makes it impossible to versions of AvaloniaEdit after this PR on older versions of Avalonia. Am I understanding correctly that if one tries to do that, there will be text misalignment problems?

danipen:
@grokys I undid the version change, so this PR can be merged now.

grokys (approved):

PR #560, opened 2026-03-05
[Potential Breaking]: Fixes multiple memory leaks and race conditions (#362)

### LINKED ISSUE #362: TextMate.Installation.Dispose does not free the TextEditor (View) for deconstruction
As is probably expected: Calling the `SetGrammer` method on a `TextMate.Installation` instance keeps the owning view from being deconstructed.

However, calling `Dispose` does not unlock it when text is already loaded in the editor.

---

## To Recreate:

- Create a new Avalonia app
- Set the output type to `Exe` to show the console
- Add the `AvaloniaEdit` and `AvaloniaEdit.TextMate` packages
- Add the `StyleInclude` to the `App.axaml`
- Create a new `UserControl` with a single `TextEditor` control (xaml) and private `TextMate.Installation` field (code-behind)
- In the constructor, set the `TextMate.Installation` field from `TextEditor.InstallTextMate` and set the grammar
- Create a public function that calls `TextMate.Installation.Dispose` on the private field
- Create a deconstructor and write something to the console to know when it runs
- In `App.OnFrameworkInitializationCompleted` set `ApplicationLifetime.MainWindow` to a new window with the content being a new instance of the custom control
- Next, cast `ApplicationLifetime.MainWindow.Content` to the custom control and call the cleanup method (can't use a variable cause that will stop it from being disposed in the function)
- Wait 1 second using `await Task.Delay(1000)` and then set `ApplicationLifetime.MainWindow.Content` to `null`
- Wait another second before calling `GC.Collect()` and `GC.WaitForPendingFinalizers()`

### Expected Result

The deconstructor will run and print to the console

### Actual Result

Nothing is printed until you comment out `SetGrammar` or set the `TextEditor.Text` to an empty string

---

A minimal sample can be found on my github - https://github.com/ArchLeaders/AvaloniaEditDisposing (check the todo list to find SetGrammar)

udlose:
Fixes #362

> [!Important]
> - The only change whose fix _could be_ a breaking change is the last memory leak - **and that is only a breaking change _if_ a user is calling `RegisterExceptionHandler`**.
> - Any callers not using `RegisterExceptionHandler` continue to work unaffected and still gain the benefit of the other fixes.
> - There are a total of:
>    - 6 or 7 leaks in the current code (depending how you group them together)
>    - 8 race conditions

## Summary

Resolves multiple memory leaks and data races in `TextMate.Installation` and `TextMateColoringTransformer` that cause the entire document object graph to be retained indefinitely after disposal.

## What was leaked per `Installation` instance

- `TextMateColoringTransformer` — retained in `LineTransformers` collection after disposal
- `TextView` ⟷ `transformer` circular reference via `VisualLinesChanged` event subscription
- `TMModel` + all per-line tokenization state
- `TextEditorModel` + `TextDocument` snapshot (potentially the entire document text)
- `ReadOnlyDictionary<string, string>` theme color map
- `Dictionary<int, IBrush>` brush cache
- `AppliedTheme` event delegate chain — roots all subscribers (e.g., ViewModels)
- Static `_exceptionHandler` — permanent GC root for the lifetime of the AppDomain if the `Action<Exception>` target captures an instance

## Impact

If applications create new `Installation` instances per document switch (rather than keeping one per editor lifetime), each switch leaks the entire previous document graph. In long-running applications with frequent document switching, this manifests as unbounded memory growth.

### Leak 1: Transformer ownership not tracked

- **Root cause:** The constructor checks `LineTransformers` for a pre-existing `TextMateColoringTransformer`. If one exists, the `Installation` uses it — but `Dispose()` unconditionally removes it from `LineTransformers` and calls `Dispose()` on it. This is a double-dispose/dangling-reference bug when the transformer was created by a prior `Installation` that still owns it.

- **Fix:** Added `_ownsTransformer` flag. The constructor records whether it created the transformer. `Dispose()` only removes/disposes the transformer when `_ownsTransformer` is `true`. When `false`, `Dispose()` calls `transformer.SetModel(null, null)` to sever the stale `_document`/`_model` references we pushed into it, without disturbing the transformer's lifecycle.

### Leak 2: Reentrancy during Dispose recreates disposed objects

- **Root cause:** `Dispose()` had no disposed-state guard. The `_editor.DocumentChanged -= OnEditorOnDocumentChanged` unsubscription in `Dispose()` could race with `OnEditorOnDocumentChanged` firing during teardown. If the event handler ran after the old objects were disposed but before unsubscription completed, it would create fresh `TextEditorModel` and `TMModel` instances assigned to fields that nobody would ever dispose.

- **Fix:** Added `_isDisposed` with `Volatile.Write` in `Dispose(bool)` and `Volatile.Read` in all other methods. `_isDisposed` is set to `true` before any teardown begins. `OnEditorOnDocumentChanged` checks `_isDisposed` under lock and silently returns if disposed, preventing reentrancy.

### Leak 3: Fields not nulled after disposal

- **Root cause:** After disposing `_editorModel`, `_tmModel`, and `_transformer`, the fields still held references to the (now-disposed) objects. Any external reference to the `Installation` (e.g., via `AppliedTheme` subscribers or ViewModel fields) roots the entire object subgraph for GC.

- **Fix:** All mutable fields are nulled under lock in `Dispose(bool)`: `_editorModel`, `_tmModel`, `_transformer`, `_grammar`, `_themeColorsDictionary`, `_exceptionHandler`, and `AppliedTheme`.

### Leak 4: `AppliedTheme` event delegate chain not cleared

- **Root cause:** If external code subscribes to `AppliedTheme` and doesn't explicitly unsubscribe, the delegate chain creates a bidirectional root between the subscriber and the `Installation`, preventing GC of either.

- **Fix:** `AppliedTheme = null` in `Dispose(bool)` under lock. The delegate reference is captured into a local before nulling (for thread-safe invocation in `SetTheme`).

### Leak 5: Transformer retains stale model/document references when not owned

- **Root cause:** During `OnEditorOnDocumentChanged`, the `Installation` calls `_transformer.SetModel(_editor.Document, _tmModel)`, which stores references to the document and TMModel inside the transformer. When `_ownsTransformer` is `false` and `Dispose()` runs, these references linger inside the transformer indefinitely.

- **Fix:** Two-part fix. (1) Added null guard in `TextMateColoringTransformer.SetModel`: `if (_grammar != null && _model != null)` — prevents NRE when `model` is `null`. (2) `Installation.Dispose(bool)` calls `transformer.SetModel(null, null)` in the non-owned path to clear stale references.

### Leak 6: Static `_exceptionHandler` creates permanent GC root

- **Root cause:** `_exceptionHandler` was a `static` field on the `TextMate` class. If the `Action<Exception>` target captures an instance (e.g., a ViewModel's error handler method), that instance and its entire reachable object graph can never be collected for the lifetime of the AppDomain.

- **Fix:** Moved `_exceptionHandler` to an instance field on `Installation`, passed via the constructor. Removed `RegisterExceptionHandler()` static method. **This is a breaking API change** — callers must now pass the handler to `InstallTextMate()` instead of calling the static method.

### Data races in `Installation`

All mutable fields in `Installation` are now synchronized via a `private readonly object _lock`. The pattern used throughout:

- **Public API methods** (`SetGrammar`, `SetGrammarFile`, `SetTheme`, `TryGetThemeColor`): `Volatile.Read` fast-path `ThrowIfDisposed()` → `lock` → authoritative `ThrowIfDisposed()` → work under lock → external calls (Redraw, event invocation) outside lock
- **Event handlers** (`OnEditorOnDocumentChanged`): `Volatile.Read` fast-path → `lock` → authoritative check → work under lock. Silent return on disposed (no exception in event handlers)
- **Dispose**: `Volatile.Read` fast-path → `lock` → `Volatile.Write(ref _isDisposed, true)` → capture field refs into locals → null fields → release lock → teardown using captured locals outside lock (prevents deadlocks from external callbacks re-entering `_lock`)
- **Read-only single-reference properties** (`EditorModel`, `TryGetThemeColor`): `Volatile.Read` without lock where the value is self-contained (no dependent fields needed atomically)

`Volatile.Write` is used for the `_isDisposed = true` assignment to pair with `Volatile.Read` callers outside the lock, satisfying analyzers (ParallelChecker) that require matching synchronization primitives on both sides of a shared variable.

### Data races in `TextMateColoringTransformer`

`ModelTokensChanged` is called from the tokenizer's **background thread** — the only cross-thread entry point. All mutable field access is now synchronized:

- **`TransformLine`** (rendering hot path): Captures 4 field references under lock (brief hold — just pointer copies), then performs all rendering work using only captured locals outside the lock. `GetLineTransformations` takes all dependencies as parameters and performs zero mutable field reads.
- **`ModelTokensChanged`**: Captures a consistent snapshot of `_model`, `_document`, `_areVisualLinesValid`, `_firstVisibleLineIndex`, `_lastVisibleLineIndex` under lock. The `Dispatcher.UIThread.Post` lambda captures **only locals** — zero mutable field reads at dispatch time, eliminating the race between the `Post()` call (background thread) and lambda execution (UI thread).
- **`SetTheme`**: Copy-on-write pattern for `_brushes`. The new dictionary is built **outside** the lock (color parsing and `ImmutableSolidColorBrush` creation are the expensive operations). The reference is swapped atomically under lock. In-flight `TransformLine` calls that captured the old dictionary continue safely — the old dictionary is never mutated, only replaced. `Dictionary.Clear()` is intentionally absent because it would corrupt concurrent readers holding captured references.
- **`_areVisualLinesValid`, `_firstVisibleLineIndex`, `_lastVisibleLineIndex`**: Changed from `volatile` to lock-protected. The lock provides the memory barriers; `volatile` was redundant overhead.

## Performance optimizations

- **`GetLineTransformations`**: Hoisted `document.GetLineByNumber(lineNumber).Offset` before the token loop. The line number is invariant across all tokens in a line, so the B-tree lookup runs once instead of once-per-token. An early return in `TransformLine` for empty/null token lists avoids both the `GetLineByNumber` call and the `ArrayPool` rent/return.
- **`SetTheme` brush construction outside lock**: Color parsing and brush allocation happen before acquiring `_lock`, minimizing lock hold time.
- **`TransformLine` lock hold minimized**: Only 4 pointer copies under lock; all rendering work is lock-free.
- **`_textView` and `_exceptionHandler` made `readonly`**: Compiler-enforced immutability eliminates synchronization requirements for these fields.

## Dispose pattern

Both `Installation` and `TextMateColoringTransformer` now implement the standard `Dispose(bool)` pattern:

- `public void Dispose()` → `Dispose(disposing: true)` + `GC.SuppressFinalize(this)`
- `protected virtual void Dispose(bool disposing)` — allows subclasses to extend teardown

`TextMateColoringTransformer` now explicitly implements `IDisposable`.

## Files changed

- **TextMate.cs**: All memory leak fixes, thread synchronization, Dispose pattern, instance-scoped `_exceptionHandler`
- **TextMateColoringTransformer.cs**: Thread synchronization, Dispose pattern, copy-on-write brushes, `SetModel` null guard, performance optimizations

### DIFF
--- a/src/AvaloniaEdit/Editing/SelectionLayer.cs
+++ b/src/AvaloniaEdit/Editing/SelectionLayer.cs
@@ -43,6 +43,11 @@ namespace AvaloniaEdit.Editing
 
         public override void Render(DrawingContext drawingContext)
         {
+            if (!_textArea.TextView.VisualLinesValid)
+            {
+                return;
+            }
+
             base.Render(drawingContext);
 
             var selectionBorder = _textArea.SelectionBorder;


danipen:
@Gillibald I'll review this but it would be cool someone in the Avalonia team to take a look too if possible. Thanks!

maxkatz6:
I might not be familiar with this code, but one potential solution, is to wrap callbacks in weak refernce, while keeping API stable:
```
private static readonly List<WeakReference<Action<Exception>>> _handlers = new();

public static void RegisterExceptionHandler(Action<Exception> handler)
{
    _handlers.Add(new WeakReference<Action<Exception>>(handler));
}

private static void Notify(Exception ex)
{
    // iterate over handlers
}
```


Or even split delegate onto handler.Target and handler.Method, and only wrap target in weak reference.

maxkatz6:
But I am good with current approach, where Installation is disposable

udlose:
> I might not be familiar with this code, but one potential solution, is to wrap callbacks in weak refernce, while keeping API stable:
> 
> ```
> private static readonly List<WeakReference<Action<Exception>>> _handlers = new();
> 
> public static void RegisterExceptionHandler(Action<Exception> handler)
> {
>     _handlers.Add(new WeakReference<Action<Exception>>(handler));
> }
> 
> private static void Notify(Exception ex)
> {
>     // iterate over handlers
> }
> ```
> 
> Or even split delegate onto handler.Target and handler.Method, and only wrap target in weak reference.

> [!Important]
> - The only change whose fix could be a breaking change is the last memory leak - and that is only a breaking change _if_ a user is calling `RegisterExceptionHandler`.
> - Any callers not using `RegisterExceptionHandler` continue to work unaffected and still gain the benefit of the other fixes.
> - There are a total of:
>    - 6 or 7 leaks in the current code (depending how you group them together)
>    - 8 race conditions

Hi @maxkatz6, @danipen.  I'm not clear if you are asking for a reconsideration of the solution or if this was a signoff on the solution implemented here.

Thanks for the feedback on the WeakReference approach. I did consider that but found issues with wrapping delegates (or their Target/Method split) in weak references. The lambdas and static methods both fail silently since their delegate Target is either a compiler-generated display class with no external strong reference, or null. The upstream `_exceptionHandler` is a single static `Action<Exception>`, not multicast. It's read once during Installation construction and passed into `TextEditorModel` and `TextMateColoringTransformer`. The static field creates a permanent GC root for the lifetime of the AppDomain if the caller's handler captures an instance. 

I can think of 2 options:
1. instance parameter only (breaking change to API). THis would involve removing the `RegisterExceptionHandler` method (the API break).
   - Pros:
      - the "breaking change" only breaks callers who have the bug - they need to change their code anyway
      - if anyone is currently unaware they have a leak, then they still get it "fixed"
      - requires users who are using `RegisterExceptionHandler` to take action to fix the leak making them aware that there was a leak
      - cleanest API, deterministic lifetime, no static root, no ambiguity
      - handler is cleaned up with `Installation` instance via `Dispose`
      - simple
   - Cons:
      - breaking change requires action/changes if caller is using `RegisterExceptionHandler`
      
2. maintain backwards compatibility with an optional instance-based exception handler. This would keep `RegisterExceptionHandler` but provide an instance-based alternative via optional param in `Installation` ctor. THe instance-based param would take precedence when provided and fallback to the static when not provided.
   - Pros:
      - non-breaking change
      - provides _option_ to fix leak via new instance-based param in `Installation` ctor param
   - Cons:
      - users who are using `RegisterExceptionHandler` would still have a massive memory leak
      - users who aren't forced to make changes are likely unaware that a memory leak is present
      - if caller uses both paths, the static one is silently ignored (instance wins). This follows the "most specific scope wins" convention, but it's implicit.
      - static handler registered after `Installation` construction won't be seen by that instance (snapshot-at-construction semantics).

I recommend Option 1 even though it _could be_ a breaking change - you are only doing it for those users who already  have a leak and you are doing so for their benefit.

PR #558, opened 2026-02-19
Migrate to avalonia 12 preview 1

latop2604:
Fix https://github.com/AvaloniaUI/AvaloniaEdit/issues/557

PR #562, opened 2026-03-09
Go to avalonia12 preview2

valerybokov:


valerybokov:
This PR is based on PR https://github.com/AvaloniaUI/AvaloniaEdit/pull/558.
I don't expect it to be merged. It's added to make the transition to Avalonia 12 Preview 3 easier when it's released. Perhaps someone will use this code as a basis.

PR #600, opened 2026-06-12
Fix: CompletionWindow closes when clicking/dragging the scrollbar thumb

### LINKED ISSUE #593: CompletionWindow closes when clicking/dragging the scrollbar thumb
**Root cause::**
CompletionList.OnPointerPressed calls RequestInsertion on any pointer press without checking whether the press source is actually a ListBoxItem. The original WPF version in icsharpcode/AvalonEdit has this guard but it was not ported to the Avalonia version.

Expected behavior:

Clicking/dragging the scrollbar scrolls the list without closing the popup.

mgarstenauer:
Fixes #593

Clicking the completion window scrollbar accidentally closes the completion window. Occurs only when

```
Editor.Options.CompletionAcceptAction = CompletionAcceptAction.PointerReleased;
```

is set.


danipen (approved):
LGTM

PR #607, opened 2026-08-07
Cap rendering and editing of extremely long lines

ollieb-unity:
### Problem

A single very long line — e.g. a minified blob, or a Unity YAML asset with an inline binary payload serialized as hex — freezes the UI thread as it scrolls into view. Two full-line passes run synchronously: TextMate tokenizes and colorizes the whole line, and `TextFormatter` shapes every character.

### Change

- **`VisualLine`** shapes only the first `LENGTH_LIMIT` characters and collapses the remainder into a short `...` marker via a `FormattedTextElement` — the same mechanism already used for collapsed fold regions. Document offsets are unchanged, so caret, selection and highlighting keep working.
- **`TextMateColoringTransformer`** skips tokenizing/colorizing lines past `LENGTH_LIMIT`.
- New **`LongLineEditProtection`** disallows edits inside the collapsed tail; **`TextArea`** routes `CanInsert` / `GetDeletableSegments` through it so callers cannot insert into or delete the hidden region (deleting the line delimiter would otherwise merge the next line into the hidden text).

This builds on the existing `VisualLine.LENGTH_LIMIT` (3000) short-circuit, which previously still added a single full-length `VisualLineText` for over-limit lines.

Builds cleanly against `net8.0` and `net10.0`.


danipen (approved):
LGTM

PR #609, opened 2026-08-15
Fix UI freeze when drag-selecting past the right edge

### LINKED ISSUE #606: Infinite UI freeze when drag-selecting past the right edge (regression from #500)
Avalonia 12.1.1, AvaloniaEdit 12.0.0 (also reproduced on current master).

Steps:
1. Open a file containing a line wider than the editor viewport (e.g. a YAML
   connection string, ~200 chars).
2. Press the left mouse button inside that line and drag past the right edge
   of the window (so the caret is placed beyond the visible area and the
   editor auto-scrolls horizontally).

Expected: selection extends and follows the auto-scroll.
Actual: the UI thread freezes indefinitely (app becomes unresponsive).

Cause (confirmed via headless reproduction + dotnet-stack dump):
infinite event loop in SelectionMouseHandler:

  TextArea_MouseMove → ExtendSelectionToMouse → Caret.BringCaretToView
    → TextView.MakeVisible → SetScrollOffset → ScrollOffsetChanged
    → TextView_ScrollOffsetChanged → ExtendSelectionToMouse(_lastMousePosition)
    → BringCaretToView → MakeVisible → SetScrollOffset → ScrollOffsetChanged → …

With the caret beyond the viewport, MakeVisible always computes a new offset,
and the ScrollOffsetChanged handler added in #500 re-runs
ExtendSelectionToMouse on the frozen mouse position, so the loop never
converges. The render/layout pass is re-triggered forever.

Reproduction: Avalonia.Headless test — TextEditor with a long line,
Window.MouseDown + repeated Window.MouseMove past the right edge with
Dispatcher.UIThread.RunJobs(); the test never completes without a fix.

Proposed fix (works in fork): re-entrancy guard — run
ExtendSelectionToMouse from TextView_ScrollOffsetChanged at most once per
PointerMoved:

    private bool _isExtendingFromScroll;

    // in TextArea_MouseMove, after updating _lastMousePosition:
    _isExtendingFromScroll = false;

    private void TextView_ScrollOffsetChanged(object sender, EventArgs e)
    {
        if (_mode is not (SelectionMode.Normal or SelectionMode.WholeWord
            or SelectionMode.WholeLine or SelectionMode.Rectangular))
            return;
        if (!TextArea.TextView.VisualLinesValid || _isExtendingFromScroll)
            return;
        _isExtendingFromScroll = true;
        ExtendSelectionToMouse(_lastMousePosition);
    }

mgarstenauer:
Fixes #606

### Problem

The UI will freeze when dragging the selection past the right edge of the text editor. See details in #606.

### Cause

The expected behavior is that the scrolling stops when the end of the scroll bar is reached. However, in this case `Caret.BringIntoView(5.0)` is called, which adds an artificial border around the caret. When this border lies outside of the range of the scroll viewer then `TextView.MakeVisible` sets an invalid, out-of-range scroll offset. This triggers another `ScrollOffsetChanged` event with the invalid scroll offset. This causes an endless loop, the drag-selecting/scrolling never stops.

### Solution

In `TextView.MakeVisible` clamp the scroll offset to the allowed range.

### DIFF
--- a/src/AvaloniaEdit/Editing/Caret.cs
+++ b/src/AvaloniaEdit/Editing/Caret.cs
@@ -459,7 +459,7 @@ namespace AvaloniaEdit.Editing
         /// </summary>
         public void BringCaretToView()
         {
-            BringCaretToView(0);
+            BringCaretToView(5.0);
         }
 
         public void BringCaretToView(double border)
--- a/src/AvaloniaEdit/Editing/SelectionMouseHandler.cs
+++ b/src/AvaloniaEdit/Editing/SelectionMouseHandler.cs
@@ -763,7 +763,7 @@ namespace AvaloniaEdit.Editing
                     TextArea.Caret.Offset = newWord.Offset < _startWord.Offset ? newWord.Offset : Math.Max(newWord.EndOffset, _startWord.EndOffset);
                 }
             }
-            TextArea.Caret.BringCaretToView(5.0);
+            TextArea.Caret.BringCaretToView();
         }
         #endregion
 
--- a/src/AvaloniaEdit/Rendering/TextView.cs
+++ b/src/AvaloniaEdit/Rendering/TextView.cs
@@ -1576,9 +1576,16 @@ namespace AvaloniaEdit.Rendering
             {
                 newScrollOffsetY = rectangle.Bottom - _scrollViewport.Height;
             }
+
             newScrollOffsetX = ValidateVisualOffset(newScrollOffsetX);
             newScrollOffsetY = ValidateVisualOffset(newScrollOffsetY);
             var newScrollOffset = new Vector(newScrollOffsetX, newScrollOffsetY);
+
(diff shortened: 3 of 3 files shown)


src/AvaloniaEdit/Editing/Caret.cs
```diff
@@ -459,7 +459,7 @@ public Rect CalculateCaretRectangle()
         /// </summary>
         public void BringCaretToView()
         {
-            BringCaretToView(0);
+            BringCaretToView(5.0);
```
mgarstenauer, line 462: In most cases only `caret.BringCaretToView()` is called, which doesn't guarantee that the caret is visible.

For example (tested in AvaloniaEdit.Demo): On a long line when I press End, the caret jumps to the end of the line. The end of the line is visible, but the caret is just slightly out of view.

⇒ I suggest to make `BringCaretToView(5.0)` the default.

Issue #206, opened 2022-02-09
Tab sign change the position of line


guilhermeolisi:
Hi,

When ShowTabs=true a sign is shown in the text but the line is shifted to the right, so tab indentation line is shifted to right relative to white space indentation line. I tested in AvalonEdit (WPF) and the line with tab is not shifted. See figures below:

AvaloniaEdit, ShowTabs=false, ShowSpaces=false 
![image](https://user-images.githubusercontent.com/25594740/153209712-503fc0db-fd2a-4b92-be40-bb7516a055ba.png)

AvaloniaEdit, ShowTabs=true, ShowSpaces=true
![image](https://user-images.githubusercontent.com/25594740/153209558-2b7e571a-3229-47af-9ee7-da1a136fa9d4.png)

AvalonEdit
![image](https://user-images.githubusercontent.com/25594740/153205881-6f4e4d9b-c619-4f41-8dd4-55338febb21f.png)

thanks for your attention

danipen:
@guilhermeolisi, fixed. Could you verify that #207 fixes the issue for you? Thanks!

guilhermeolisi:
> @guilhermeolisi, fixed. Could you verify that #207 fixes the issue for you? Thanks!

Hi @danipen, how can i verify #207? Do I need to download the source code of the control and test it? Sorry, I'm not a professional developer and I'm not familiar with this

danipen:
Yes you can clone the repository and reference it from your csproj instead of referencing the Avalonia nugets.

danipen:
Alternatively ... maybe @Takoooooo may create a new release. I'm that case you'd only have to update the AvaloniaEdit version in you csproj.

danipen:
Yes I saw. I'm not aware about how line wrapping is implemented in AvaloniaEdit, sorry. 

Let's see if someone else knows how to fix that issue.

guilhermeolisi:
> Yes I saw. I'm not aware about how line wrapping is implemented in AvaloniaEdit, sorry.
> 
> Let's see if someone else knows how to fix that issue.

Thanks @danipen , i will wait then

Gillibald:
As far as I remember WordWrap isn't implemented yet

danipen:
@Gillibald wrapping is partially implemented. I will continue the conversation in #205

PR #611, opened 2026-08-27
Align tabs to the next tab stop instead of a fixed advance

ollieb-unity:
## Problem

Text after a tab does not line up on a tab stop.

`TabTextElement` emits `TextCharacters("\t")` and leaves the width to the formatter, but the text shaper gives every `\t` glyph a constant advance (`TextShaperOptions.IncrementalTabWidth`, see `src/Skia/Avalonia.Skia/TextShaperImpl.cs` in Avalonia) with no knowledge of the column the tab starts at. A tab therefore always advances `IndentationSize` spaces instead of reaching the next multiple of it, so text after a tab only lines up when it happens to start on a tab stop already.

This shows up in any file that uses tabs to align trailing comments or declarations. Taking eight declarations whose tab counts make the comments align at tab size 4, here is the column each comment lands on:

```text
with tab stops (expected):  28,28,28,28,28,28,28,28
with a fixed advance (now): 29,30,31,28,29,30,31,28
```

The comments drift one column per line and snap back every fourth line. Leading indentation looks correct either way, which is why this is easy to miss: a tab at column 0, 4, 8 lands on a stop under both behaviours.

## Change

- `SingleCharacterElementGenerator` measures the column where the tab starts, expanding any earlier tabs on the same line, and emits the spaces needed to reach the next multiple of `IndentationSize`, rather than handing `"\t"` to the formatter.
- `TabTextElement.VisualLength` now varies with that count, and `CreateTextRun` handles a run request that lands part way into the element. That happens when the line wraps, and the old two column element could not express it.
- The zero width `TabGlyphRun` is untouched, so the fix from #207 / #206 (tab glyph extra width) still holds. The arrow is still drawn as an overlay at the tab's start and contributes no width.
- Tabs are laid out whether or not `ShowTabs` is set, since the width was wrong in both cases. Consequently `TextView` now always attaches the generator: it owns tab layout and could otherwise be dropped when `ShowSpaces`, `ShowTabs` and `ShowBoxForControlCharacters` are all off.

## Test

`TextViewTests.Tab_Should_Reach_The_Next_Tab_Stop` asserts the rendered runs for three cases at `IndentationSize = 4`: a tab from column 1 and one from column 3 both reach column 4, and a tab already sitting on a stop reaches the next one. It asserts run text rather than pixel positions, so it does not depend on font metrics.

Without the source change the test fails with `Expected: "a   b" But was: "a\tb"`. The full suite passes with it (516 tests).

## Two things worth your opinion

1. **This might belong in Avalonia rather than here.** If the shaper applied `IncrementalTabWidth` as a tab stop instead of a constant advance, the original `TextCharacters("\t")` would be correct and this change would be unnecessary. I fixed it in AvaloniaEdit because AvaloniaEdit already owns a `TabTextElement` and decides how tabs render, but I am happy to move it upstream into Avalonia if you would rather have it there.
2. **The `TextView` change alters default behaviour** for every consumer, since the generator becomes always present. It is needed for the fix to apply when the whitespace glyphs are hidden. Happy to split it into its own commit or drop it if you would prefer to keep tabs broken in that configuration for now.


Gillibald:
IncrementalTabWidth is a fixed width for the tab character. This does not align anything to a specific column layout. This is working as intended. For column layouts, there needs to be an array of tab stops so one can define a width individually.

danipen:
Thanks @Gillibald yeah I think the fix belong to this project and we should assume that AvaloniaEdit uses monospaces font (same as VSCode), so I think this is the right approach.

@ollieb-unity thanks for the PR. I reviewed it, and I think we can make a couple of improvements:
- I think we can extract the tab rendering and the calculation into a dedicated generator and extract it from the `SingleCharacterElementGenerator`. So `SingleCharacterElementGenerator` still can be registered under demand (no changes there).
- Also I think we can easily track the DisplayColumn in the VisualLine so we can calculate later the position in the generator O(1) per tab, instead of O(n).
- We also can improve the mem handling in the generator, and instead of using `new string(' ', spaceCount)` we can have just one `IndentationSize-length` string, and peek the slices we neededwith `AsMemory(0, n)`

I will handle those changes. I think I have a clear idea.

danipen:
I've opened #612 as a rewrite with the changes I mentioned above.

